PR #31, opened 2018-01-04
Fix alignment issues and column overflow

Niharika29:
Things in this commit:
* Repositioning of the settings and delete icons. Having them next to the program/event name makes it clearer on what the icons are for. 
* Center align columns. 
* Show a message instead of an empty table when there are no events in a program.
* Add a max-width for columns to avoid the issue seen below-
Before:
gmetrics-broken
After:
gmetrics-fixed

I think we should also add a character limit for the program and event titles.


app/Resources/views/programs/index.html.twig
```diff
@@ -19,13 +19,13 @@
...
-                    <th>{{ msg('day-retention', [30]) }}</th>
-                    <th>{{ msg('pages-created') }}</th>
-                    <th>{{ msg('pages-improved') }}</th>
                     <th></th>
+                    <th class="center-align-cell">{{ msg('events') }}</th>
+                    <th class="center-align-cell">{{ msg('participants') }}</th>
+                    <th class="center-align-cell">{{ msg('new-editors') }}</th>
+                    <th class="center-align-cell">{{ msg('day-retention', [30]) }}</th>
+                    <th class="center-align-cell">{{ msg('pages-created') }}</th>
+                    <th class="center-align-cell">{{ msg('pages-improved') }}</th>
```
MusikAnimal: Since we're only using `center-align-cell` to center text, we might could use the Boostrap `text-center` class instead of creating a new rule.

app/Resources/assets/css/application.scss
```diff
@@ -31,7 +31,16 @@ body {
...
 
 .table th {
-    font-weight: normal;
+    font-weight: bold;
+    max-width: 250px;
+}
+
+.table td {
+    max-width: 250px;
+}
```
MusikAnimal: Here we could instead make every column except the first have `white-space: nowrap`, then do away with the `max-width`. That way larger screens can take advantage of the extra space and not wrap at 250px. For this I'd use the `text-nowrap` class on all but the first `<th>` element.

Niharika29: Isn't that default? It won't wrap and I'm trying to do that. If I do away with max-width and add text-nowrap class to columns besides the first, it'd do nothing effectively. How can I tell it to wrap after a limit without specifying max-width for columns? I could do the max-width in % if you think that's better?

app/Resources/views/programs/index.html.twig
Niharika29: Done.

app/Resources/assets/css/application.scss
MusikAnimal: If `text-nowrap` is on all the other cells (`<td>` and `<th>`), then the title should wrap if there's not enough room, looking similar to your second screenshot. I suppose you might run into a scenario where the headings are really close together (though you could add padding), depends on what you want. Just a recommendation, the `max-width` will indeed make sure things don't look bad so I've no problem with it :)

Niharika29: I tried this. It didn't wrap but overflowed outside the viewport. 
screen shot 2018-01-10 at 10 46 03 am

Gonna push the code now so you can try it too.

I know `max-width` isn't a really neat solution but I couldn't find any other that'd work well.

app/Resources/views/programs/index.html.twig
```diff
@@ -19,22 +19,22 @@
...
+                    <th class="text-center text-nowrap">{{ msg('new-editors') }}</th>
+                    <th class="text-center text-nowrap">{{ msg('day-retention', [30]) }}</th>
+                    <th class="text-center text-nowrap">{{ msg('pages-created') }}</th>
+                    <th class="text-center text-nowrap">{{ msg('pages-improved') }}</th>
                 </tr>
             </thead>
             <tbody>
                 {% for program in programs %}
-                    <tr class="program-entry">
+                    <tr class="program-entry text-nowrap">
```
MusikAnimal: Trying remove this `text-nowrap`, since we want the title to wrap, and not all the cells including the title.

Niharika29: That worked! Silly me, I totally missed that. I messed up while rebasing this to fix conflicts so now it think the autocompletion changes were introduced in this PR. :/ I fixed the other stuff though.
I added another tiny change to not show an empty table if there are no events in a program and instead show a message saying that.

MusikAnimal:
Weird that the already-merged changes are showing up as part of this PR. Going to close and create another one... hold tight!

+1 to your other change. Empty tables aren't pretty!

MusikAnimal:
OK it's now at #33

PR #33, opened 2018-01-10
Fix alignment issues and column overflow

MusikAnimal:
Tweak text-wrap; Don't show empty table for no events in program

Fix alignment issues and column overflow


MusikAnimal:
@Niharika29 Merged! I always use `git pull --rebase origin master` and resolve conflicts that way (fix files, then `git add . && git rebase --continue` as necessary), but the more ideal thing would be for me to not touch code that you were actively working on :)

PR #35, opened 2018-01-17
Pages created and pages improved statistics

MusikAnimal:


src/AppBundle/Repository/EventRepository.php
```diff
@@ -51,26 +53,85 @@ public function getNumNewEditors(Event $event)
...
+     */
+    public function getNumPagesEdited($dbName, DateTime $start, DateTime $end, $usernames)
+    {
+        $start = $start->format('YmdHis');
+        $end = $end->format('YmdHis');
+
+        $conn = $this->getReplicaConnection();
+        $rqb = $conn->createQueryBuilder();
+
+        $rqb->select('COUNT(page_title)')
```
Niharika29: COUNT(DISTINCT page_title)

Niharika29: Since the only difference between the Created and Edited queries is that we check for rev_parent_id = 0 in Edited and not in Created - could we make use of the Edited result set to get the Created stats? Do you think it could be a performance improvement?

MusikAnimal: Good catch on the `DISTINCT`! And yeah, we might be able to combine the Created and Edited queries somehow. I've got something in mind that might work.

PR #72, opened 2018-05-18
Add dewiki to Scrutinizer wiki farm

MusikAnimal:


MusikAnimal:
No longer needed now that we can test against the replicas in the CI build

PR #74, opened 2018-07-13
Don't remove orphan wikis when generating stats

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T199289

### DIFF
--- a/src/AppBundle/Model/EventWiki.php
+++ b/src/AppBundle/Model/EventWiki.php
@@ -39,7 +39,6 @@ class EventWiki
      */
     const FAMILY_NAMES = [
         'wikipedia',
-        'commons',
     ];
 
     /**
@@ -159,7 +158,9 @@ class EventWiki
         $family = $this->getFamilyName();
 
         return $this->event->getWikis()->filter(function (EventWiki $wiki) use ($family) {
-            return substr($this->domain, 2) === $family && $wiki->getDomain() !== $this->domain;
+            return $wiki->isChildWiki() &&
+                $wiki->getFamilyName() === $family &&
+                $wiki->getDomain() !== $this->domain;
         });
     }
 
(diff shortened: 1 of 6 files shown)


src/AppBundle/Model/EventWiki.php
```diff
@@ -159,7 +158,9 @@ public function getChildWikis()
         $family = $this->getFamilyName();
 
         return $this->event->getWikis()->filter(function (EventWiki $wiki) use ($family) {
-            return substr($this->domain, 2) === $family && $wiki->getDomain() !== $this->domain;
+            return $wiki->isChildWiki() &&
+                substr($this->domain, 2) === $family &&
```
mooeypoo: You should be careful with this; we have Wikipedias whose language code is more than 2 characters, for example,  ksh.wikipedia and the more outrageously-long one, http://zh-min-nan.wikipedia.org

(see more here https://en.wikipedia.org/wiki/List_of_Wikipedias)

Maybe split the string by periods, and check the first piece?

src/AppBundle/Service/EventProcessor.php
```diff
@@ -457,7 +457,7 @@ private function createOrUpdateEventStat($metric, $value, $offset = null)
     /**
      * Persist an EventWikiStat with given metric and value, or update the
      * existing one, if present.
```
mooeypoo: This might be unrelated, but if you change the parameter from "EventWikiStat" to "EventWiki" does that affect the purpose of the method? The documentation says "persiste an EventWikiStat..." -- should that change to "EventWiki" instead?

src/AppBundle/Model/EventWiki.php
MusikAnimal: There's actually a `EventWiki::getFamilyName()` method that I should be using here. Fixed!

src/AppBundle/Service/EventProcessor.php
MusikAnimal: This method takes an EventWiki and creates and persists an EventWikiStat for it. I've updated the doc block to make this more clear.

MusikAnimal:
This has been deployed to stating -- https://tools.wmflabs.org/grantmetrics-test/

PR #77, opened 2018-07-23
Show statistics broken down by wiki family on event page

MusikAnimal:
Some refactoring of related Twig templates.

Remove unused 'event_valid' field from event table, instead doing
validation in the model layer with Event::isValid().

Add message indicating event is missing required options before
statistics can be generated.

Change font to Roboto, and remove unused font file.

Bug: https://phabricator.wikimedia.org/T192579


i18n/en.json
```diff
@@ -55,6 +55,9 @@
 	"event-data": "Event data",
 	"event-data-for": "Event data for \"$1\"",
 	"event-deleted": "The event '$1' was deleted.",
+	"event-in-future": "Statistics can't be generated because this event takes place in future.",
```
mooeypoo: "... takes place in the future" (missing 'the'?)

src/AppBundle/Model/Event.php
```diff
@@ -221,6 +215,18 @@ public function getCacheKey()
         return (string)$this->id;
     }
 
+    /**
+     * Is the Event valid? If false, statistics will be able to be generated.
```
mooeypoo: If false, statistics will *not* be able to be generated...

src/AppBundle/Model/Event.php
```diff
@@ -221,6 +215,18 @@ public function getCacheKey()
...
+     * @return bool
+     */
+    public function isValid()
+    {
+        return $this->wikis->count() > 0 &&
+            $this->start !== null &&
```
mooeypoo: Ther'es also 'event-in-future' which says that the statistics can't be done because the event is in the future -- do we need to add that test in the validity here?

MusikAnimal:
@mooeypoo Thanks for identifying all those careless errors! =p All fixed.

MusikAnimal:
GitHub makes the first line of the commit message the title of the PR, and the rest as the description. Not sure if you saw the whole thing? f529e9c5d19c60efe9dc534dcc76eb7f72256a88 said:
```
Show statistics broken down by wiki family on event page

Remove unused 'event_valid' field from event table.

Add message indicating event is missing required options before
statistics can be generated.

Change font to Roboto, and remove unused font file.

Bug: https://phabricator.wikimedia.org/T192579
```

but I just forced pushed back, adding even more info :) Now it's at a6a85a953c44766b493ab3586cbdcd557db0ef65

mooeypoo:
... I think I was looking at an earlier one? In any case, we can also edit the top comment on the PR so that when/if we look at it on Github it's all there.

PR #83, opened 2018-08-08
Move new-participants textarea above list of existing participants

samwilson:
But still list the invalid participants above the textarea. This
change also changes the participants form from being a horizontal
form (in Bootstrap parlance) and so adds .row divs where required;
this means some styling can be removed.

Bug: https://phabricator.wikimedia.org/T200544

### DIFF
--- a/app/Resources/views/events/_participants.html.twig
+++ b/app/Resources/views/events/_participants.html.twig
@@ -7,64 +7,67 @@
         {# Valid participants should be shown below invalid ones. #}
         {% set validParticipants = [] %}
 
-        {{ form_start(form, {'method': 'post', 'attr': {'class': 'form-horizontal save-participants-form', 'autocomplete': 'off'}}) }}
-        {% for participant in form.participants %}
-            {% set invalidParticipant = false %}
-            {% if not(participant.vars.valid) %}
-                {% set invalidParticipant = true %}
-            {% endif %}
-
-            {% set participantRow %}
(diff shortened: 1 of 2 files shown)


app/Resources/views/events/_participants.html.twig
```diff
@@ -7,25 +7,29 @@
         {# Valid participants should be shown below invalid ones. #}
         {% set validParticipants = [] %}
 
-        {{ form_start(form, {'method': 'post', 'attr': {'class': 'form-horizontal save-participants-form', 'autocomplete': 'off'}}) }}
+        {{ form_start(form, {'method': 'post', 'attr': {'class': 'save-participants-form', 'autocomplete': 'off'}}) }}
+        <div class="row">
+            {# This form comprises two columns, the first for the participants list and fields, the 2nd for the submit button. #}
+            <div class="col-sm-6">
+
         {% for participant in form.participants %}
```
MusikAnimal: Indent?

MusikAnimal:
One comment, otherwise looks and works great!

Niharika had some [feedback](https://phabricator.wikimedia.org/T200544#4488703). I think I might agree that the invalid usernames appear below the textarea, but at the top of the list. If it's too much work let's not worry about it, though.

app/Resources/views/events/_participants.html.twig
samwilson: Fixed.

PR #85, opened 2018-08-09
Remove duplicate usernames after normalizing

samwilson:
Re-order the validation process of participant usernames.

Bug: https://phabricator.wikimedia.org/T201577


MusikAnimal:
Thanks!

PR #86, opened 2018-08-09
[WiP] Fix Code Climate integration

samwilson:
Just investigating this:

> $ ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
> Error: could not find any viable formatter. available formatters: clover, cobertura, coverage.py, excoveralls, gcov, gocov, jacoco, lcov, simplecov

And also the badge URLs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ script:
 
 after_script:
   - ps -ef | grep communitytech@tools-dev.wmflabs.org | awk 'NR==1{print $3}' | xargs kill -9
-  - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
+  - ./cc-test-reporter after-build --coverage-input-type clover --exit-code $TRAVIS_TEST_RESULT
 
 branches:
   only:
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ A Wikimedia Foundation tool that provides grantees a simple, easy to use interfa
 
 [![License: GPL v3](https://img.shields.io/badge/License-GPL%20v3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
 [![Build Status](https://travis-ci.org/wikimedia/grantmetrics.svg?branch=master)](https://travis-ci.org/wikimedia/grantmetrics)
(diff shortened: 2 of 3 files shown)


samwilson:
These badge URLs are not correct, I think, but I don't have access to the CodeCoverage admin area.

.travis.yml
```diff
@@ -37,11 +37,11 @@ before_script:
 script:
   - composer lint
   - php phpDocumentor.phar -d src -t docs/api --template="checkstyle" # Different than `composer docs` for CI build.
-  - composer unit
+  - ./vendor/bin/phpunit --coverage-clover ./clover.xml tests
```
MusikAnimal: I don't know if it matters, but we're using Symfony's `simple-phpunit` with `composer test` (`composer unit` specifically). Indeed we weren't generating the clover.xml file which obviously is needed!

samwilson: Oh, good point. Fixed.

samwilson:
And can you grab the badge URLs from Code Climate?

MusikAnimal:
> These badge URLs are not correct, I think

Yeah but if we keep showing Code Climate's scores they will always be good, hehe =p

> I don't have access to the CodeCoverage admin area

I don't see where to add people :/ I would try logging in with GitHub at https://codeclimate.com/login (the "Quality" version of Code Climate, not "Velocity"), then add Grant Metrics. Maybe it'll just let you in, since we're both part of the Wikimedia organization.

> And can you grab the badge URLs from Code Climate?

https://api.codeclimate.com/v1/badges/53b3ddf0ce654055f8cd/maintainability
https://api.codeclimate.com/v1/badges/53b3ddf0ce654055f8cd/test_coverage

samwilson:
> Maybe it'll just let you in, since we're both part of the Wikimedia organization.

I tried, no joy. Doesn't matter though, I don't think there's anything to be modified there now.

I've updated this patch; hopefully all good to go now.

MusikAnimal:
Thanks! What's odd was I was seeing test coverage show up on the badge up until the past few commits. Based on the URL it must have been the test coverage of the Code Climate repo, which apparently disappeared because it now just shows "?"

samwilson:
Deus ex machina I reckon, cause now it's showing 96% again. :)

PR #87, opened 2018-08-09
Create general-use panel macro with a toggling visual caret

samwilson:
Generalize the panel that currently is only being used for event
participants, to remove event-specific stuff and also add a
visual caret to the header to indicate more clearly that the
panel can be collapsed.

Bug: https://phabricator.wikimedia.org/T200551


app/Resources/assets/css/application.scss
```diff
@@ -219,3 +219,25 @@ a.disabled {
...
+.panel {
+    .panel-heading {
+        display: flex;
+        align-items: end;
+        .panel-title {
+            font-size: 24px;
+            flex: 3;
+        }
+        .panel-description {
+            flex: 1;
```
MusikAnimal: Only thing here is the text isn't vertically aligned. I accomplished this by setting an explicit `height` and `line-height` https://github.com/wikimedia/grantmetrics/pull/87/files#diff-ff1c57142581996a6cd7e089b450dc9fL112 so we might restore that.

There may be a better way to do it, though!

samwilson: I was going for the `end` alignment. But now you mention it, `center` is better! :) And no need for explicit height.

Oh, and it turns out that I'm confused: the [W3C]( https://www.w3.org/TR/css-flexbox-1/#propdef-align-items) says different things to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items). Anyway, center is what we want yeah?

samwilson:
Don't merge yet; I'm adding the tooltip to the panel title.

app/Resources/assets/css/application.scss
MusikAnimal: Yeah vertically centered, right-alighted.

samwilson: Done. (Sorry, I did the gerrit-world thing of modifying an existing commit, so GH probably didn't tell you about it.)

samwilson:
Ready for review now.

samwilson:
I forgot the qqq messages. Done now.

MusikAnimal:
Looks fantastic!

PR #88, opened 2018-08-09
Add category support to queries for pages created/improved

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T194707

### DIFF
--- a/src/AppBundle/Repository/EventCategoryRepository.php
+++ b/src/AppBundle/Repository/EventCategoryRepository.php
@@ -7,6 +7,7 @@ namespace AppBundle\Repository;
 
 use AppBundle\Model\EventCategory;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Query\QueryBuilder;
 
 /**
  * An EventCategoryRepository supplies and fetches data for the EventCategory class.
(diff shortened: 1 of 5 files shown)


src/AppBundle/Repository/EventCategoryRepository.php
```diff
@@ -26,57 +27,101 @@ public function getEntityClass()
...
+    {
+        $rqb = $this->getReplicaConnection()->createQueryBuilder();
+        $rqb->select(['DISTINCT(cl_from)'])
+            ->from("$dbName.categorylinks")
+            ->join("$dbName.categorylinks", "$dbName.page", null, 'cl_from = page_id')
+            ->where('page_namespace = 0')
+            ->andWhere(
+                $rqb->expr()->in(
+                    'cl_to',
+                    "SELECT cat_title FROM $dbName.category WHERE cat_id IN (:categoryIds)"
```
samwilson: I'm probably not thinking about this correctly, but do we need to do this with a subquery? Can we join categorylinks to both category and page, e.g.

```php
    $rqb->select(['DISTINCT(cl_from)'])
            ->from("$dbName.categorylinks")
            ->join("$dbName.categorylinks", "$dbName.category", 'cl', 'cl_to = cat_title')
            ->join("$dbName.categorylinks", "$dbName.page", 'p', 'cl_from = page_id')
            ->where('page_namespace = 0')
            ->where( 'cat_id IN (:categoryIds)'
```

tests/AppBundle/Command/ProcessEventCommandTest.php
```diff
@@ -270,4 +271,39 @@ private function jobFinishedSpec()
             ]);
         static::assertEquals(0, count($jobs));
     }
+
+    /**
+     * Creates a new job, this time with EventCategory's on the Event.
```
samwilson: no apostrophe needed

src/AppBundle/Repository/EventCategoryRepository.php
MusikAnimal: I believe you are correct!

tests/AppBundle/Command/ProcessEventCommandTest.php
MusikAnimal: PHPStorm complains `EventCategorys` is a spelling error, but I'll just add it to the dictionary :) I didn't want to put `EventCategories` because that isn't the class name.

samwilson: Yeah I'm never sure what to do in that situation either. I think though that the `ies` is probably correct, but it's clear either way. Just not the possessive! :)

PR #89, opened 2018-08-10
Fix countable error for PHP 7.2

samwilson:
Doctrine\ORM\EntityRepository::findOneBy() returns NULL if nothing
is found, and as of PHP 7.2 this isn't countable.

Also add PHP 7.2 to CI, for future bug-catching.

I'm not sure if we want to bother about 7.2... but i'm using it locally, so this was bugging me.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,7 +1,7 @@
 build:
     environment:
         php:
-            version: 5.6
+            version: 7.2
     tests:
         override:
             # Disables tests. We're only using Scrutinizer for inspections.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 
 php:
-  - '7.0'
+  - '7.2'
 
 env:
   global:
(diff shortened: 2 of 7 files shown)


MusikAnimal:
> I'm not sure if we want to bother about 7.2... but i'm using it locally, so this was bugging me.

We do want to bother! https://phabricator.wikimedia.org/T201437

Though it appears PHP 7.2 isn't preinstalled on Travis. I think it might have been a fluke, where a few of the 7.2 builds were realllly slow. This last go around it was reasonably fast, but there are some [outstanding failures](https://travis-ci.org/wikimedia/grantmetrics/jobs/414405528).

I'm not going to install PHP 7.2 until T201437 is resolved and we're ready to make the transition, but in the meantime I'm certainly OK putting testing against 7.2 on Travis. We might even remove 7.0. I somewhat worry about the two builds running parallel and using the database credentials.

samwilson:
simple-phpunit is [failing](https://travis-ci.org/wikimedia/grantmetrics/jobs/416627397#L1204) with PHP 7.2, but the same tests pass when just using phpunit directly. This seems to be something to do with how phpunit-bridge switches between phpunit 5 & 6 depending on what version of PHP is in use.

MusikAnimal:
I see. Honestly I don't even know what the advantage is to simple-phpunit! I just went with it per Symfony recommendation. I'm OK with plain ole PHPUnit, if you are :)

PR #91, opened 2018-08-13
Use timezone-adjusted start date when checking if stats can be generated

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T201739

### DIFF
--- a/app/Resources/views/events/show.html.twig
+++ b/app/Resources/views/events/show.html.twig
@@ -31,7 +31,7 @@
             </h1>
         </div>
 
-        {% if event.isValid and event.start <= date() %}
+        {% if event.isValid %}
             <div class="pull-right text-right">
                 {% if isOrganizer %}
                     <button type="button" class="btn btn-primary event-process-btn{% if jobStatus is not null %} disabled{% endif %}" data-event-id="{{ event.id }}">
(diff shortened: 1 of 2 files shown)


app/Resources/views/events/show.html.twig
```diff
@@ -31,7 +31,7 @@
             </h1>
         </div>
 
-        {% if event.isValid and event.start <= date() %}
+        {% if event.isValid and event.startWithTimezone <= date() %}
```
samwilson: `Event::isValid()` already checks that start is before now, so I think the 2nd part of this check can be removed. However, `Event::isValid()` doesn't take into account TZ, so the startWithTimezone should maybe be moved there?

app/Resources/views/events/show.html.twig
```diff
@@ -104,7 +104,7 @@
             <div class="text-center event-wiki-stats--empty">
                 {% if event.numParticipants == 0 %}
                     {{ msg('event-misconfig', [msg('participants')]) }}
-                {% elseif event.start > date() %}
+                {% elseif event.startWithTimezone > date() %}
```
samwilson: Should this be `event.isValid`?

samwilson: Sorry, I realise now this is correct as it is.

PR #92, opened 2018-08-14
Add favicon and logo to interface

MusikAnimal:
Remove unused font file

Bug: https://phabricator.wikimedia.org/T187854


MusikAnimal:
> Do we need to bump the asset version for this?

Yes we do! It would be neat to automate this somehow, but probably not worth our while since we'll be moving to Webpack at some point.

samwilson:
Hurrah for logo!

PR #94, opened 2018-08-16
Add support for pages create or improved on Wikidata

samwilson:
This adds stats for Wikidata, counting the same stats as for
Wikipedia but renaming them.

Bug: https://phabricator.wikimedia.org/T190456


i18n/en.json
```diff
@@ -91,8 +91,10 @@
 	"pager-older-n": "{{PLURAL:$1|Older $1|Older $1}}",
 	"pages-created": "Pages created",
 	"pages-created-desc": "A content page is a page in either the main or file namespace.",
+	"pages-created-wikidata": "Items created",
 	"pages-improved": "Pages improved",
 	"pages-improved-desc": "A content page is a page in either the main or file namespace.",
+	"pages-improved-wikidata": "Items improved",
```
MusikAnimal: don't forget qqq :)

samwilson: Good point! Should we think about adding a check for that to the CI? A codesniff should be added for it to Intuition maybe?

MusikAnimal: That would be magnificent! If you know how to do it, by all means :) I wasn't sure if you could easily program in your own sniff rules like that.

MusikAnimal: That would be magnificent! If you know how to do it, by all means :) I wasn't sure if you could easily program in your own sniff rules like that.

samwilson:
This might be ready for review, if you've got a mo.

MusikAnimal:
Overall looks good! There's no test coverage for `EventProcessor::setItemsCreatedOrImprovedOnWikidata()`, but I guess that's okay because it's exactly the same as `setItemsCreatedOrImprovedOnWikidata()`. On that note, should we re-use the same method? Or add `@codeCoverageIgnore` (with an explanatory comment) to the Wikidata method so our test coverage isn't impacted.

Another question: I don't know Wikidata very well... are we sure only items (starting with `Q`) are in the mainspace?

samwilson:
Good point about tests! I totally did start there, but got a tiny bit lazy. Done now.

And yep, all mainspaces pages on WD are items.

PR #96, opened 2018-08-21
Remove high unicode characters from model titles

samwilson:
This replaces 4-byte utf8 characters (i.e. emojis, basically) in
model titles with the unicode replacement character.

Bug: https://phabricator.wikimedia.org/T201388

### DIFF
--- a/src/AppBundle/Model/Traits/TitleUserTrait.php
+++ b/src/AppBundle/Model/Traits/TitleUserTrait.php
@@ -51,6 +51,10 @@ trait TitleUserTrait
      */
     public function setTitle($title)
     {
+        // Remove 4-byte unicode characters (replace with the "replacement character" �).
+        // Kudos https://stackoverflow.com/a/24672780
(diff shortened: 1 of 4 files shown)


src/AppBundle/Model/Traits/TitleUserTrait.php
```diff
@@ -51,6 +51,10 @@ public function getDisplayTitle()
      */
     public function setTitle($title)
     {
+        // Remove 4-byte unicode characters (replace with the "replacement character" �).
+        // Kudos https://stackoverflow.com/a/24672780
+        // More info http://unicode.org/reports/tr36/#Deletion_of_Noncharacters
+        $title = preg_replace('/[\x{10000}-\x{10FFFF}]/u', "\xEF\xBF\xBD", $title);
```
aezell: If we know we might need to do this with more strings in the future, would we be better off with a function available broadly that returns the corrected string?

PR #97, opened 2018-08-21
Require ext-redis in composer.json

MusikAnimal:
Update doctrine/coding-standard and phpunit


MusikAnimal:
@samwilson What do you think of this? After the PHP 7.2 upgrade I had to install ext-redis on my machine for the "revision browser" to work (where redis is used). I made this PR to add it as a dependency but apparently Travis doesn't have ext-redis, yet master is on PHP 7.2 and passing. So maybe it's a Mac thing? Should I abandon this PR or make Travis install ext-redis?

PR #95, opened 2018-08-16
[WIP] Category support on frontend; mild refactor of EventController

MusikAnimal:

PR #98, opened 2018-08-21
API endpoint to get all wikis adhering to EventWiki::VALID_WIKI_PATTERN

MusikAnimal:
This will be used in clientside JavaScript for autocompletion on wiki
inputs. We can't go by the Sitematrix API because newly introduced wikis
may be production but not yet on the replicas, causing Grant Metrics to
error out.

This commit also adds ext-json as a dependency in composer.json, along
with some unrelated minor code aesthetics.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         ]
     },
     "require": {
+        "ext-json": "*",
         "php": "^5.6|^7.0",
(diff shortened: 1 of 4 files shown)


MusikAnimal:
Merging this to clean up #95. We're gonna want this new method either way.

PR #100, opened 2018-08-22
[WIP] Add CategorySubscriber to handle title/ids; add some getters/setters/counters on Event and EventWiki

MusikAnimal:
**This pull request is a part of #95**

Add CategorySubscriber to set title/ids on ORM load/persist

Add a getter and counter of categories on an Event.

Add a dbName property to EventWiki, so that it can be cached and
used when running queries.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -34,6 +34,11 @@ services:
             - { name: doctrine.event_listener, event: postLoad }
             - { name: doctrine.event_listener, event: prePersist }
 
+    AppBundle\EventSubscriber\CategorySubscriber:
+        tags:
+            - { name: doctrine.event_listener, event: postLoad }
+            - { name: doctrine.event_listener, event: prePersist }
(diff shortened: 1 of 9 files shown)


MusikAnimal:
Sorry I should have put this as WIP, since per [T202762#4543378](https://phabricator.wikimedia.org/T202762#4543378) we're probably going to be storing titles and not IDs. I'll move the str_replace and trim logic to EventCategory as suggested, with a `getTitle()` (underscored, for database queries) and `getDisplayTitle()` (no underscores), which matches TitleUserTrait. We might still want the CategorySubscriber to populate the dbname on the associated EventWikis, maybe... since the dbname is needed when validating the category.

MusikAnimal:
Aborting. Whole new PR coming!

MusikAnimal:
NOTE: Plz don't delete this branch yet :)

PR #103, opened 2018-08-28
Fix variable naming in revision wikitext export

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T202707


MusikAnimal:
Marking as WIP as I want to add integration test for the wikitext and CSV export actions.

PR #105, opened 2018-08-29
Upgrade to Symfony 4

samwilson:


samwilson:
Not sure why `"public-dir": "web"` isn't working as expected.

PR #106, opened 2018-08-29
Put prod assets in version control

MusikAnimal:
This is recommended by Symfony, and also will get around the
Node issues we are running into on Toolforge.

See https://symfony.com/doc/current/frontend/encore/faq.html#do-i-need-to-install-node-js-on-my-production-server

Bug: https://phabricator.wikimedia.org/T202471


aezell:
What triggers the rebuilding of the asset files? Are they built on every build in Travis (or whatever)? If so, how are the filenames generated? Do the filenames change if the contents of the file hasn't changed?

I'm wondering what happens if you commit certain filenames and then another engineer commits a different set of filenames. There'll potentially be a git merge conflict but maybe the file contents haven't even changed.

MusikAnimal:
Development is done in app/Resources/assets. You specify the files in [webpack.config.js](https://github.com/wikimedia/grantmetrics/blob/master/webpack.config.js#L20). Then when the `./node_modules/.bin/encore production` command is ran it compiles and concatenates them into app.js and app.css. These two files are versioned automatically, and sourced with [`{{ asset('assets/app.js') }}`](https://github.com/wikimedia/grantmetrics/blob/285075322a549f2071088a03d6364a80e1f14645/app/Resources/views/base.html.twig#L25) in the `<head>`. So there could be filename conflicts but the same was true with Assetic.

I guess we have to remember to compile the production assets in our pull requests, but similarly we had to remember to bump the asset version when we were using Assetic. Travis doesn't care about assets because the integration tests use a headless browser. That being said it would be nice if only the deploy script compiled production assets, and we didn't have to commit them, but we shall bow to convention (and we have Node issues on Toolforge anyway).

MusikAnimal:
> I guess we have to remember to compile the production assets in our pull requests

Actually this I guess isn't a problem, so long as we always compile as `production` and not `dev` when working locally (since you'll obviously want to compile assets one way or another in order to see your changes during development). Your browser (at least Chrome, Firefox, & Safari) has an un-minify feature, so you could still easily add in JavaScript breakpoints when debugging.

aezell:
I like the idea of the pre-commit hook.

PR #107, opened 2018-08-30
Move web directory in preparation for Symfony 4.

samwilson:
The default public directory will be 'public/' so we could just
add another symlink (or change the existing one) but it seems
neater to reorganise things so we only need one docroot.


MusikAnimal:
This did not work on my local. I had to symlink public to point to public_html.

The public_html thing is specific to Toolforge. We probably won't stay on Toolforge forever, so if you want I'm okay with a symlink for the time being.

samwilson:
Okay, good point. Let's just rename the actual dir to `public` (to match the new Symfony standard) and update the symlink to point to that.

I'm not sure why the `symfony-web-dir` in composer.json doesn't work (it changes to `public-dir` in Symfony 4 anyway). But I've left it in; maybe something's amiss on my local.

MusikAnimal:
Yeah `symfony-web-dir` didn't work for me either, it seems. But the current PR does, tested on staging :) I guess once we're on VPS or a production instance we won't need to override the web/public directory anyway. Merging!

PR #109, opened 2018-09-01
Store titles in EventCategory instead of IDs

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T202762


src/AppBundle/Model/Event.php
```diff
@@ -365,17 +372,67 @@ public function setTimezone($timezone)
...
+     * @param EventWiki $wiki
+     * @return string[]
+     */
+    public function getCategoryTitlesForWiki(EventWiki $wiki)
+    {
+        return $this->categories->filter(function (EventCategory $category) use ($wiki) {
+            return $category->getDomain() === $wiki->getDomain();
+        })->map(function (EventCategory $category) {
+            return $category->getTitle();
+        })->toArray();
```
aezell: My PHP is admittedly rusty but this code is very confusing to me. What is it doing?

MusikAnimal:
> The only thing I can think of is that there's now the theoretical possibility for a category to exist with a domain that isn't an EventWiki. Does this matter? It makes it simpler in other ways, for sure.

Definitely possible, and we are doing a little "data redundancy", but the simplicity I think trumps any other technical quirks. We will definitely need to validate the domain, making sure that it matches EventWikis configured on the event.

samwilson:
Sounds sensible. Go for it I say.

src/AppBundle/Model/Event.php
MusikAnimal: For starters, this is using Doctrine's [ArrayCollection](https://www.doctrine-project.org/api/collections/latest/Doctrine/Common/Collections/ArrayCollection.html) to traverse the arrays, effectively the object-oriented style of `array_filter` and `array_map`. Not sure if that was the confusing part. On a high level, this method first gets categories on the event that are configured for the given wiki, then loops through again to get their titles. I'll add some comments for clarity.

PR #110, opened 2018-09-04
Update Alice to most recent version

samwilson:
I'm not quite sure this is the fully best way, but it works and doesn't resort to the extra dependency that I think is the proper way. :)

Bug: https://phabricator.wikimedia.org/T203139


MusikAnimal:
Interesting that the tests aren't failing in master because of the changes to the method signature of `EventWiki::__construct`? Anyway, all looks good now. Merging!

samwilson:
Yeah, I wasn't quite sure what was going on there either. :man_shrugging:

PR #111, opened 2018-09-04
Move Intuition to be a service, and inject it where required

samwilson:
Bug: https://phabricator.wikimedia.org/T203421

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,7 +2,7 @@ version: "2"        # required to adjust maintainability checks
 checks:
     argument-count:
         config:
-            threshold: 4
+            threshold: 5
     complex-logic:
         config:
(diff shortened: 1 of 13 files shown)


samwilson:
This is not complete yet. The outstanding part is to resolve the `EntityController::validateUser()` by maybe moving it to a [before event filter](https://symfony.com/doc/current/event_dispatcher/before_after_filters.html) (at the moment, it's called from the Controller constructor, which is no good because the Intuition service isn't injected until after then).

MusikAnimal:
> This may be a good opportunity to fix how we use Intuition with flash messages.

Now that I read the Phab task you created, you're clearly aware of this! Ha. I'll reply there

samwilson:
Yep, all passes now.

PR #112, opened 2018-09-05
Improve phpdoc installation and configuration

samwilson:
Temporarily removing phpdoc checkstyle from Travis CI because it
wasn't checking correctly anyway (was always successful).

Moved other phpdoc config into phpdoc.dist.xml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,13 @@
 /public/fonts/
 /output/
 /docs/api/
+/public/docs/
+/build/api-cache/
 /.idea/
 
 *.swp
 .DS_Store
 i18n_helper.js
 phpDocumentor.phar*
+/bin/phpdoc*
+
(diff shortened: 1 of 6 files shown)


samwilson:
Current status: PhpDocumentor 2 doesn't support PHP 7.2, so we need to upgrade to 3. But 3 is still in alpha, and has some bugs. Like, seemingly, not caring about some options (like title), and not being able to produce a checkstyle report.

So I'm not really sure what we should do. I think it's worth tidying up `composer docs` so that it can generate docs, but not use it in CI at all. If there are docblock rules we want to enforce, we can probably add them to phpcs anyway.

Ideas?

aezell:
Maybe this is blasphemy but I think I'm OK leaving it disabled until 3 is ready.

MusikAnimal:
Yeah, I guess there's not much we can do. If you are able to add things to phpcs, and it's not too much trouble, go for it. Otherwise we'll just have to rely on code review for the time being. IDEA/PhpStorm complains about the lack of docs, at least, and we're both using it.

PR #113, opened 2018-09-06
Update dependencies and improve linting

samwilson:
This updates all dependencies and adds extra linting to CI.


samwilson:
This leaves `sensio/framework-extra-bundle` in, even though I don't think we're actually using any of it now that `@Route` annotations are in Symfony core, because it seems likely that we'll want to do e.g. `@Template` annotations at some point. Or not; we could take it out till then. I just wanted to clean up the deprecation notices... :-)

PR #114, opened 2018-09-06
Add ISO8601 override for English date formatting

samwilson:
Also add US English as an option, so that the old default can
still be selected if desired.


src/AppBundle/Service/Intuition.php
```diff
@@ -40,6 +40,10 @@ public static function serviceFactory(RequestStack $requestStack, SessionInterfa
         $intuition = new static('grantmetrics');
         $intuition->registerDomain('grantmetrics', $rootDir.'/../i18n');
         $intuition->setLang(strtolower($useLang));
+
+        // Also add US English, so we can access the locale information (e.g. for date formatting).
+        $intuition->addAvailableLang('en-us', 'US English');
```
MusikAnimal: I would probably call it `'English (US)'` or is there some standard I'm unaware of? Language first would make it easier to find in the language dropdown.

PR #117, opened 2018-09-11
Add config for custom sendmail config (for Toolforge)

samwilson:
This is just an idea I've got about making it email us critical errors.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -144,3 +144,32 @@ toolforge:
         logged_in_user: '%app.logged_in_user%'
     intuition:
         domain: 'grantmetrics'
+
+monolog:
+    handlers:
+        main:
+            type: fingers_crossed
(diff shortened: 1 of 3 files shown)


samwilson:
hm, seems there's more to it in the php container:

>  Process could not be started [sh: 1: /usr/sbin/exim: not found]

MusikAnimal:
Fantastic idea!! I don't know how the mailer things work, but we should do this, if it's not incredibly difficult.

samwilson:
Yeah, I've got it working locally. The trouble on toolforge is that there's no MTA in the PHP container, and I don't know if we've got a SMTP server we can use (actually, maybe we do?).

MusikAnimal:
I would ask in #wikimedia-cloud on IRC. Bryan certainly would know.

PR #120, opened 2018-09-13
Add help link to new-participants textbox label

samwilson:
Bug: https://phabricator.wikimedia.org/T192749


app/Resources/assets/css/events.scss
```diff
@@ -103,15 +103,13 @@
     }
 
     .save-participants-form {
+        .new-participants-label label:after {
```
MusikAnimal: Psuedoelement should be prefixed with two colons per W3C spec: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Syntax

app/Resources/views/events/_participants.html.twig
```diff
@@ -8,11 +8,17 @@
         {% set validParticipants = [] %}
 
         {{ form_start(form, {'method': 'post', 'attr': {'class': 'save-participants-form', 'autocomplete': 'off'}}) }}
+        <p class="new-participants-label">
+            <label for="form_new_participants">{{ msg('add-more-participants') }}</label>
+            <a href="https://meta.wikimedia.org/wiki/Special:MyLanguage/Grant_Metrics_tool#{{ msg('add-more-participants-help-link-fragment') }}"
```
MusikAnimal: I just [added](https://meta.wikimedia.org/w/index.php?title=Grant_Metrics_tool&diff=18376905&oldid=18370889) an `{{anchor}}` template at the top of the section, with the ID `adding-participants`. This way we can link to it and it will work regardless of the language. Otherwise I guess the translators would need to check and see what the section is, and of course it could change at any time.

samwilson: Good idea! I have no idea what I was thinking. :-)

Fixed.

app/Resources/assets/css/events.scss
samwilson: I learn something every day.

Fixed.

PR #115, opened 2018-09-06
Frontend for category support; refactor form handling in EventController

MusikAnimal:
Add migration to make event_category unique on event, title and domain, not just title and domain.

Remove invalid categories upon saving an Event.

Rework EventWikiRepository::getDbName to go off of a domain and not an EventWiki object, since EventCategory is not tied to an EventWiki.

Note #121 is a subset of this PR. That one can be reviewed and merged first to make reviewing this one a bit easier.

PR #121, opened 2018-09-17
Refactor add/remove, autocompletion and other event-related JavaScript

MusikAnimal:
This simplifies how grantmetrics.application.setupAddRemove() and
setupAutocompletion() work, and sets up the ability for other kinds
of lists (e.g. categories) to use them.

Break out event-related JS into separate files for each view.

Add better commenting for all JS methods, and include example usage.


MusikAnimal:
This is part of #115

PR #122, opened 2018-09-18
Switch to ToolforgeBundle

samwilson:

PR #123, opened 2018-09-18
Make PHP classes strict; use scalar type hints and return types

MusikAnimal:
This takes advantage of PHP 7.2 scalar type hints and return types.
Some things will likely break, but those things are likely bugs
which is why I think doing this is a good idea.

See https://symfonycasts.com/screencast/php7/scalar-type-hints for more.

Other various cleanup.


MusikAnimal:
I did not change Intuition-related methods and classes knowing they would conflict with #122

PR #124, opened 2018-09-26
Use IDs in routes instead of titles; disallow IDs in titles

MusikAnimal:
Because we're disallowing IDs in titles (and there are no existing
events/programs with numerical titles), we can go off of is_numeric
to make the routing backwards-compatible.

Bug: https://phabricator.wikimedia.org/T192748


i18n/qqq.json
```diff
@@ -49,6 +49,7 @@
 	"error-server-message": "Error message. $1 is the message given by the server.",
 	"error-service-overload": "Error message shown when too many people are using Grant Metrics.",
 	"error-title-invalid-chars": "Error message shown when creating a program or event with a title containing invalid characters. $1 is a list of the characters that can't be in the title.",
+	"error-title-numeric": "Error message shown when creating a program or event with a title that is a number (does not contain any non-numerical characters).",
```
aezell: Maybe this should say "with a title that is only a number..."?

src/AppBundle/Controller/EntityController.php
```diff
@@ -81,41 +81,57 @@ private function setProgramAndEvent(): void
...
-            }
+        $programId = $this->request->get('programId');
+
+        if (!$programId) {
+            return;
+        }
+
+        $repo = $this->em->getRepository(Program::class);
+
+        if (is_numeric($programId)) {
```
aezell: Do you want to test for an integer? `is_numeric` will give `True` for floats.

src/AppBundle/Controller/EntityController.php
```diff
@@ -81,41 +81,57 @@ private function setProgramAndEvent(): void
...
-            }
+        $eventId = $this->request->get('eventId');
+
+        if (!$eventId) {
+            return;
+        }
+
+        $repo = $this->em->getRepository(Event::class);
+
+        if (is_numeric($eventId)) {
```
aezell: Same thing here about integers.

src/AppBundle/Model/Traits/TitleUserTrait.php
```diff
@@ -75,6 +75,20 @@ public function validateUnreservedTitle(ExecutionContext $context): void
...
     }
 
+    /**
+     * Validates that the title is not just a number (as these would be assumed to be IDs).
+     * @Assert\Callback
+     * @param ExecutionContext $context
+     */
+    public function validateNonNumerical(ExecutionContext $context): void
+    {
+        if (is_numeric($this->title)) {
```
aezell: Another spot where maybe `is_int` would be better.

src/AppBundle/Controller/EntityController.php
aezell: Perfect! I was right and wrong at the same time. Pretty typical.

That function does look like what we need. It even throws false if handed an integer.

MusikAnimal:
I'm assuming this is okay to merge; the only thing that changed since the last :+1: was adding tests and removing some now-unneeded code.

samwilson:
Yep, sorry I didn't make it back to re-review.

PR #125, opened 2018-09-27
Add most Slevomat coding standards, most notably return types

MusikAnimal:
See https://github.com/slevomat/coding-standard

Most of these sniffs should be uncontroversial, and not only help
maintain more consistent code, but also prevent bugs (especially
requirement to use return types and type hinting).

Almost all sniffs can be automatically fixed with phpcbf.


composer.json
```diff
@@ -21,21 +21,22 @@
...
         "ext-redis": "*",
-        "php": "^7.2",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/doctrine-migrations-bundle": "^1.0",
         "doctrine/orm": "^2.5",
         "eightpoints/guzzle-bundle": "^7.0",
         "leafo/scssphp": "^0.7.1",
         "sensio/framework-extra-bundle": "^5.2",
-        "symfony/webpack-encore-pack": "^1.0",
+        "slevomat/coding-standard": "^4.8",
```
samwilson: Should this be a dev requirement?

MusikAnimal:
(merged manually)

> We could also think about adding this coding standard to the bundle, and using the ruleset from there instead of defining it all in this phpcs.xml.

That makes sense to me! We're using most of the Slevomat standards so it might be easier to include everything and then blacklist what we don't want. I assume the local app can override things too in phpcs.xml

PR #130, opened 2018-10-05
Add 'bdi' Twig function, and use it for user-generated text

mooeypoo:
Wrap all user-generated content in bdi tags, so that directionality
is preserved even when read through other interface languages.

More specifically, event and program names in Farsi, Arabic or
Hebrew (RTL) will be desplayed correctly even if the user
viewing them in English (LTR)

**
Depends on merging the pull request on ToolforgeBundle first. 
https://github.com/wikimedia/ToolforgeBundle/pull/6
**

See Phabricator task: https://phabricator.wikimedia.org/T205172

### DIFF
--- a/app/Resources/views/events/show.html.twig
+++ b/app/Resources/views/events/show.html.twig
@@ -9,10 +9,10 @@
             </li>
         {% endif %}
         <li>
-            <a href="{{ path('Program', {'programId': program.id}) }}">{{ program.displayTitle }}</a>
+            <a href="{{ path('Program', {'programId': program.id}) }}">{{ bdi(program.displayTitle) }}</a>
         </li>
         <li class="active">
-            {{ event.displayTitle }}
+            {{ bdi(event.displayTitle) }}
         </li>
     </ol>
 {% endblock %}
@@ -21,7 +21,7 @@
     <div class="container">
         <div class="page-header">
             <h1>
-                {{ event.displayTitle }}
+                {{ bdi(event.displayTitle) }}
                 {% if isOrganizer %}
                     <small>
                         (<a href="{{ path('EditEvent', {'programId': program.id, 'eventId': event.id}) }}">{{ msg('edit-event')|lower }}</a>)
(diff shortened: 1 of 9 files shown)


mooeypoo:
I accidentally created a duplicate pull request rather than amending the previous one. Sorry about that. Refer to #127 for historical review.

This one has the helper methods out, and only contains fixes to the .twig files. The methods that do the actual bdi wrapping are now in https://github.com/wikimedia/ToolforgeBundle/pull/6

app/Resources/views/events/show.html.twig
```diff
@@ -21,7 +21,7 @@
     <div class="container">
         <div class="page-header">
             <h1>
-                {{ event.displayTitle }}
+                {{ bdi(event.displayTitle) }}</bdi>
```
MusikAnimal: I'm guessing the `</bdi>` shouldn't be there?

mooeypoo: Whoops that's a remnant of an experiment. Thanks for catching that!

mooeypoo:
Rebased the patch, as well.

I could not test this locally -- my new computer has some annoying issues running grantmetrics that I'm still slowly solving. Please test this before approving/merge, just in case.

MusikAnimal:
I think we need to run `composer update wikimedia/toolforge-bundle` to update the composer.lock file. I will give this a go assuming you're still having issues with your dev environment

MusikAnimal:
There are still a few places that needed the `bdi()` Twig function ([example](https://github.com/wikimedia/grantmetrics/blob/master/app/Resources/views/programs/show.html.twig#L71)), which is easy enough to fix.

I'm not sure what to do about flash messages, though:

screen shot 2018-10-15 at 17 48 24

It seems we can't wrap the whole message in `<bdi>`. The message itself is [constructed in PHP](https://github.com/wikimedia/grantmetrics/blob/master/src/AppBundle/Controller/EventController.php#L92), and I don't think HTML goes there... what to do?

MusikAnimal:
I've added in some missing `bdi()`s and am merging this. We can revisit the flash messages later.

PR #137, opened 2018-10-17
Update links to the project page on Meta

MaxSem:


MusikAnimal:
The pages are about to get moved to "Event Metrics", so I would wait on this. Even if we don't rename the app, I think the docs should be at "Grant Metrics" and not "Grant Metrics tool"

MusikAnimal:
Links have all been updated to the new Event Metrics page on Meta. Thanks for this PR, anyway!

PR #138, opened 2018-10-18
Only add a blank category to the CategoriesType form

MusikAnimal:
Before we were adding it for all forms, so submitting the
ParticipantsType form would attempt to save the blank category,
bypassing form-level validations, and would error out.


aezell:
This change seems sane. If it works correctly on `grantmetrics-test` we should deploy to production.

PR #140, opened 2018-11-07
 Add a stat for number of edits during an event

MaxSem:


MaxSem:
WIP: can't get this to actually output the number, and it needs tests.

MusikAnimal:
One thing you may have missed -- we need to make the new `edits` metric available to all wikis (since it's applicable to all of them), so you'll want to add it to the `*` array in [Event.php](https://github.com/wikimedia/grantmetrics/blob/master/src/AppBundle/Model/Event.php#L64). Still, I'm not sure we want to show the number in the interface, yet. Let's check with Joe. If we do want it hidden (and only in the reports that we build later), I'll take care of [T207909](https://phabricator.wikimedia.org/T207909) ASAP, unless you are interested in doing it.

For the tests, see the [fixtures](https://github.com/wikimedia/grantmetrics/blob/master/src/AppBundle/DataFixtures/ORM/extended.yml#L24-L39) for the data we're working with. I think for the "Oliver and Company" (a classic movie, by the way!) we're looking at 9 total edits (`EventStat`), and you'll want to test the values for maybe at least one `EventWikiStat` too. Using `pages-improved` as an example:
* [ProcessEventCommandTest.php](https://github.com/wikimedia/grantmetrics/blob/master/tests/AppBundle/Command/ProcessEventCommandTest.php#L183-L204)
* [EventTest.php](https://github.com/wikimedia/grantmetrics/blob/master/tests/AppBundle/Model/EventTest.php#L362)

Code otherwise looks great! We're missing the `qqq` documentation for the `edits` i18n message, but with a follow up commit I'm going to import the identical `edits` message and its translations from XTools.

MusikAnimal:
Thanks! Right now Edits are showing up in the UI, but we don't have to deploy to prod just yet. I'm going to work on [T207909](https://phabricator.wikimedia.org/T207909) tomorrow.

PR #142, opened 2018-11-14
Rename application to 'Event Metrics'

MusikAnimal:


app/Resources/views/base.html.twig
```diff
@@ -5,9 +5,9 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <title>
             {% if gmTitle is defined %}{{ msg_if_exists(gmTitle) }} - {% elseif exception is defined %}{{ msg('fatal-error') }} - {% endif %}
-            {{ msg('grant-metrics') }}
+            {{ msg('event-metrics') }}
         </title>
-        <link rel="icon" type="image/x-icon" sizes="16x16 32x32 48x48" href="{{ asset('favicon.ico') }}" />
+        <link rel="icon" type="image/x-icon" sizes="16x16 32x32 48x48" href="{{ asset('favicon.ico') }}?v=2" />
```
samwilson: Cache-busting could be handled by webpack like other assets are, if this file was in the assets directory and something `require()`'d it.

Probably a separate issue though.

i18n/en.json
```diff
@@ -62,6 +63,8 @@
 	"event-data-for": "Event data for \"$1\"",
 	"event-deleted": "The event '$1' was deleted.",
 	"event-in-future": "Statistics can't be generated because this event takes place in the future.",
+	"event-metrics": "Event Metrics",
+    "event-metrics-welcome": "Welcome to Event Metrics",
```
samwilson: Indenting awry?

app/Resources/views/base.html.twig
samwilson: e.g. here: https://github.com/wikimedia/svgtranslate/blob/59c41227cefe02cf7b7e77b5b60e84e21b3c1e18/assets/app.js#L11

PR #143, opened 2018-11-14
Use APCu cache adapter instead of Redis

MusikAnimal:
Bug: T208818


aezell:
I'm in love with this PR. I will try to test/review it later.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -264,19 +264,17 @@ public function getRevisions(Event $event, ?int $offset = 0, ?int $limit = 50, b
...
         // in which case we'll want to invalidate the cache.
         $shouldUseCache = null !== $event->getUpdated() &&
-            (int)$event->getUpdated()->format('U') < time() - $cacheDuration &&
-            null !== $this->getRedisConnection();
+            (int)$event->getUpdated()->format('U') < time() - $cacheDuration;
         $cacheKey = $this->getCacheKey(func_get_args(), 'revisions');
-        if ($shouldUseCache && $this->getRedisConnection()->contains($cacheKey)) {
-            return $this->getRedisConnection()->fetch($cacheKey);
+        if ($shouldUseCache && $this->cache->hasItem($cacheKey)) {
+            return $this->cache->getItem($cacheKey)->get();
```
aezell: According to [this doc](https://api.symfony.com/4.1/Symfony/Component/Cache/Simple/ApcuCache.html), there is a simpler `ApcuCache`. Looks like we could use the simpler `get($key...)` and `set($key...)`. Are there performance differences between that and the full [`ApcuAdapter`](https://api.symfony.com/4.1/Symfony/Component/Cache/Adapter/ApcuAdapter.html) which does provide `getItem()`?

src/AppBundle/Repository/EventRepository.php
```diff
@@ -264,19 +264,17 @@ public function getRevisions(Event $event, ?int $offset = 0, ?int $limit = 50, b
...
         }
 
         $ret = $this->getRevisionsData($event, $offset, $limit, $count);
 
         // Cache for 5 minutes.
         if ($shouldUseCache) {
-            $redis = $this->getRedisConnection();
-            $redis->save($cacheKey, $ret, $cacheDuration);
+            return $this->setCache($cacheKey, $ret, 'PT5M');
```
aezell: This method [`setCache`](https://api.symfony.com/4.1/Symfony/Component/HttpFoundation/Response.html#method_setCache) appears to be related to a `Response` object and not really the `Cache` system. Maybe we want something like [`save()`](https://api.symfony.com/4.1/Symfony/Component/Cache/Adapter/ApcuAdapter.html#method_save) instead?

MusikAnimal: It probably could have a better method name, but this is from [Repository.php#L188](https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/Repository/Repository.php#L188). It caches whatever you give it for the specified time, and returns the same thing (`mixed` type). So basically it's just a shorthand.

src/AppBundle/Repository/EventRepository.php
MusikAnimal: It's actually using https://symfony.com/doc/current/components/cache/cache_pools.html, the PSR-6 implementation that allows you to change the caching backend without having to update any code.

MusikAnimal:
I'm gonna self-merge because Alex is out, and this is holding back use of the new VPS instances. In particular product/QA can't test the "edit list" page (formerly "revision browser") on eventmetrics-dev.

If further changes are needed I will promptly apply them with a subsequent PR :)

PR #144, opened 2018-11-19
Use constant to define what metrics should be shown in the interface

MusikAnimal:
This hides the "Edits" metric that was added with #140, bringing the
interface to what it is currently in production.

This commit does not use the VISIBLE_METRICS constant in the
downloadable reports and wikitext export.

Bug: https://phabricator.wikimedia.org/T207909


MusikAnimal:
You can ignore the Scrutinizer report.

MusikAnimal:
Thanks!

PR #145, opened 2018-11-20
Redesign of navigation and headers of pages

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T206495


MusikAnimal:
Reviewer note: All those i18n changes were removals of now-obsolete messages, and importing translations for "settings" and "save" from another application (where they are used in an identical fashion). So basically you can just review en.json and qqq.json :)

mooeypoo:
This looks good to me overall. A couple of notes:

1. Seems Joe discovered something missing https://phabricator.wikimedia.org/T206495#4776203
2. I think it would be good if, in the future, we split apart the language import part of these. The first review is then the code + en.json/qqq.json, and the followup immediately after merge can be all the changed language files from the migration script. It will make review (and general cleanliness) a lot cleaner and streamlined.

Not a blocker for this PR, but could be good to do for next time.

I'm waiting to merge this for the fix of the first bullet point --- otherwise, seems good to be merged.

mooeypoo:
Things seem to be fixed; whatever followup can be done in a followup PR.

PR #146, opened 2018-11-21
Start background process to spawn jobs; e.g. real job queue

MusikAnimal:
Remove code that returns statistics to the views from the process event
endpoint, since this isn't being used.

Bug: https://phabricator.wikimedia.org/T192954


tests/AppBundle/Controller/EventDataControllerTest.php
```diff
@@ -168,28 +168,25 @@ public function testStats(): void
         $this->response = $this->client->getResponse();
         static::assertEquals(404, $this->response->getStatusCode());
 
+        /**
+         * FIXME: Find a way to retain data across multiple processes in functional tests.
+         * Because the process event endpoint starts a background process, PHPUnit will truncate all data
+         * that was created earlier in the test. Hence, the app:spawn-job command will fail because the Job
+         * was erased. The code below was from before background processing was implemented.
```
mooeypoo: I think I am misunderstanding something with the processes, but this strikes me as similar to general async code tests. 

Laravel has a "queue" functionality that can be mocked in tests (see https://laravel.com/docs/5.7/mocking#queue-fake ) and since Laravel is built on Symfony, I wonder if there's anything we can use here. 

Moreover, I am wondering -- can we not use async-style testing on this? If not, will something like this not work?

```
$process->start();
$process->wait();
static::assertEquals( .... );
```

MusikAnimal: That FIXME was from an earlier implementation, I just forgot to remove it. However as it turns out I was still doing it wrong. The important thing we want here is to create the Job, and attempt to start it immediately (if there's quota), as opposed to waiting for the cron to start it which could be up to 60 seconds later. Also, we want to start it in the background so the user can browse to another page, etc. I have managed to figure this out by using a [listener for the kernel.terminate event](https://github.com/wikimedia/eventmetrics/pull/146/files#diff-a03fda7eed4e489eca7b27603175f4e8R189).

Unfortunately this true "background" process system is untestable in a single-threaded server like your local (unless you use Apache or whatever). Instead it does everything synchronously, so the tests will pass. I think this okay... but yourself or QA might want to help test things once deployed to VPS. At the time of writing it's on staging https://eventmetrics-dev.wmflabs.org

With this PR I've also added an [endpoint to check the status of a job](https://github.com/wikimedia/eventmetrics/pull/146/files#diff-a03fda7eed4e489eca7b27603175f4e8R207), and I've made the event page [poll for the job status](https://github.com/wikimedia/eventmetrics/pull/146/files#diff-0e925f7438fa188ad78f88033099da74R66) so you get real-time updates.

With that I think we're all set as far as having a real job queue system.

MusikAnimal: > With this PR I've also added an endpoint to check the status of a job, and I've made the event page poll for the job status so you get real-time updates.

FYI, this also means on your local it always says "The statistics are waiting to be generated..." even though it's already started. That's because it can't make a separate request to get the job status, since it's tied up processing the event. Once it's done though it will refresh the page.

mooeypoo: Does our CI test this properly? within the VPS? 

In general the PR looks good to me, I'm just a little confused (still) about the tests. If nothing can test the tests, can we change them, or can we make sure we have a proper coverage in case this is not working? It is a pretty big feature we need to make sure works.

mooeypoo: Answered in chat -- this *is* tested even if not completely async, which I think is enough to make sure the feature is operational for the long term with CI tests.

MusikAnimal:
Thanks for the review! :)

PR #148, opened 2018-11-27
Remove trailing slash routes and fix tests

MusikAnimal:
In Symfony 4 trailing slashes automatically redirect to the same route
without the slash. In Symfony 4.1.9 something else changed that caused
our manually added trailing slash routes to fail.

This commit removes all trailing slash routes and fixes the integration
tests accordingly.

### DIFF
--- a/app/config/parameters-travis.yml.dist
+++ b/app/config/parameters-travis.yml.dist
@@ -36,9 +36,8 @@ parameters:
     oauth.key:                  ~
     oauth.secret:               ~
 
-    # Path from domain to get to the root of the app. On Toolforge for instance this is /eventmetrics
-    # Do not include a trailing slash.
-    app.root_path: ''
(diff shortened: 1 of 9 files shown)


mooeypoo:
Looks good overall, but a test fails:

```
1) Tests\AppBundle\Controller\EventControllerTest::testLoggedOut
LogicException: The request was not redirected.
/home/travis/build/wikimedia/eventmetrics/vendor/symfony/symfony/src/Symfony/Component/BrowserKit/Client.php:533
/home/travis/build/wikimedia/eventmetrics/tests/AppBundle/Controller/EventControllerTest.php:132
ERRORS!
```

MusikAnimal:
This only fails on Travis. Still investigating

MusikAnimal:
Merging based on Moriel's review; the only that changed since then was I set the `app.root_url` parameter for the Travis parameters.yml, which is why the test was failing only on Travis.

PR #149, opened 2018-12-11
Bump Composer packages

MaxSem:
Hardcode Symfony version to avoid an upgrade to 4.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "symfony/monolog-bundle": "^3.3",
         "symfony/polyfill-apcu": "^1.0",
         "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/symfony": "^4.0",
+        "symfony/symfony": "4.1.9",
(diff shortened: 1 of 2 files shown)


MusikAnimal:
@MaxSem #148 was merged. Did you want to try upgrading to Symfony 4.2? Or is there another reason we're avoiding it?

MaxSem:
My main concern was that it needs moar testing, but now that you're saying it's fine...

MusikAnimal:
Tests are pretty comprehensive, so I have faith things will be mostly OK if they pass. Symfony also claims everything is backwards-compatible, where you would have received warnings before needing any code changes https://symfony.com/blog/symfony-4-2-curated-new-features As far as I know we've no warnings running on 4.1.8

MusikAnimal:
Everything seems to run fine in 4.2, but indeed there are numerous depredations that look stressful... 4.1.9 it is, for now :)

PR #150, opened 2018-12-12
Make 'Update data' say 'Updating' when updating

MusikAnimal:
With #146 we made the job queue act like a real job queue (backgrounding
the job process). I forgot to fix the JavaScript so that once the job
starts, the "Update data" button says "Updating". Right now it always
says "Queued" until it has finished.

With this PR we're also updating node dependencies, to fix a security
vulnerability. This included an upgrade to Encore itself, which now
requires we source runtime.js as per deprecation. Webpack-encore seems
to have this new system documented only at:
https://github.com/symfony/webpack-encore/blob/master/index.js#L507-L528

Bug (broadly related): https://phabricator.wikimedia.org/T192954


mooeypoo:
I'm a little confused with this change, a couple of quick clarifying questions --
* Is this connected to a ticket? If so, please add it in?
* Can we split out the updating of the node dependencies from the fix that adds "update" to update when updating? It's very confusing otherwise.
* Let's be a little more descriptive in the commit description+issue description.

MusikAnimal:
It is confusing! I have updated the commit message to better explain what's going on.

As for the node dependencies -- On my fresh new Ubuntu machine I hadn't ran `npm install` yet, which is needed so I can compile the assets using Encore. When I did this I got a warning about a vulnerability, so I ran `npm audit fix`, which in turn updated a bunch of node dependencies. This ballooned into accommodating the new version of Encore which requires you put `.enableSingleRuntimeChunk()` (or `disable`), along with sourcing the new runtime.js file.

I often include updates to dependencies as part of a PR, if and only if those updates are required. Also as you know GitHub doesn't have any "depends on" functionality... so maybe it's okay to keep them bundled, assuming the commit summary is informative enough?

mooeypoo:
I think what threw me off was the vague PR/commit message, this is much better -- thank you!

I would have preferred that at least the security vulnerability thing be in its own commit, even if it's in the same PR, even just for the sake of git-blame / git-bisect operations in case we ever need them in the future.

In this one, I'm okay moving forward -- but I am going to try and think if there's a good way to do this in a sustainable git-world-way while not stopping anyone's workflow while developing. The mashing stuff together can be disruptive in the future when and if we examine past commits or go over history.

Ideas are welcome on that one, considering the limitations of github...

MusikAnimal:
Merging based on Moriel's comment above. I finally fixed the Travis build. I had to make it use a newer version of Node.

PR #152, opened 2018-12-13
Add pageviews stats

samwilson:
Bug: https://phabricator.wikimedia.org/T206817

### DIFF
--- a/src/AppBundle/Model/Event.php
+++ b/src/AppBundle/Model/Event.php
@@ -49,6 +49,8 @@ class Event
         'byte-difference' => null,
         'pages-created' => null,
         'pages-improved' => null,
+        'pages-created-pageviews' => null,
+        'pages-improved-pageviews' => 30,
         'files-uploaded' => null,
         'file-usage' => null,
         'items-created' => null,
@@ -60,7 +62,13 @@ class Event
      */
     public const WIKI_FAMILY_METRIC_MAP = [
         '*' => ['edits', 'new-editors', 'retention'],
-        'wikipedia' => ['pages-created', 'pages-improved', 'byte-difference'],
+        'wikipedia' => [
+            'pages-created',
+            'pages-improved',
+            'byte-difference',
+            'pages-created-pageviews',
+            'pages-improved-pageviews',
+        ],
         'commons' => ['files-uploaded', 'file-usage'],
         'wikidata' => ['items-created', 'items-improved'],
     ];
--- a/src/AppBundle/Repository/EventWikiRepository.php
+++ b/src/AppBundle/Repository/EventWikiRepository.php
@@ -9,6 +9,7 @@ namespace AppBundle\Repository;
 
 use AppBundle\Model\Event;
 use AppBundle\Model\EventWiki;
+use DateInterval;
 use DateTime;
 use Doctrine\DBAL\Connection;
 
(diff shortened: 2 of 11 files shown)


MaxSem:
What's left to be done here?

i18n/en.json
```diff
@@ -109,6 +109,10 @@
 	"pages-created-desc": "A content page is a page in either the main or file namespace.",
 	"pages-improved": "Pages improved",
 	"pages-improved-desc": "A content page is a page in either the main or file namespace.",
+	"pages-created-pageviews": "Views to pages created",
+	"pages-created-pageviews-desc": "Total cumulative views to pages created in the main namespace.",
+	"pages-improved-pageviews": "Avg. daily views to pages improved",
+	"pages-improved-pageviews-desc": "Average daily views over the preceding 30 days to pages improved during the event.",
```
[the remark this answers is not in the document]
samwilson: Actually, as these aren't used in this patch, I've pulled them out; they can be added by the patch that adds the reports.

(we should make the banana checker work some time I guess...)

src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -257,6 +258,62 @@ public function getPageIds(
...
+    ): int {
+        if (0 === count($pageIds)) {
+            return 0;
+        }
+        $conn = $this->getReplicaConnection();
+        $rqb = $conn->createQueryBuilder();
+        $rqb->select('page_title')
+            ->from("$dbName.page")
+            ->where('page_id IN ('.implode(',', $pageIds).')');
+        $stmt = $this->executeQueryBuilder($rqb);
```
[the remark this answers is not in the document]
samwilson: Oops, good point. I didn't see that; much better idea.

Done.

src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -257,6 +258,56 @@ public function getPageIds(
...
+        EventWiki $wiki,
+        DateTime $start,
+        array $pageIds,
+        bool $getDailyAverage = false
+    ): int {
+        if (0 === count($pageIds)) {
+            return 0;
+        }
+        $pageviewsRepo = new PageviewsRepository();
+        $recentDayCount = 30;
```
MaxSem: This should be defined in Event::AVAILABLE_METRICS

src/AppBundle/Repository/PageviewsRepository.php
```diff
@@ -0,0 +1,97 @@
...
+     * Given a Mediawiki article and a date range, returns a daily timeseries of its pageview counts.
+     * @link https://wikimedia.org/api/rest_v1/#!/Pageviews_data/get_metrics_pageviews_per_article_project_access_agent_article_granularity_start_end
+     * @param EventWiki $eventWiki
+     * @param string $article Page title with underscores. Will be URL-encoded.
+     * @param string $granularity The time unit for the response data, either GRANULARITY_DAILY or GRANULARITY_MONTHLY.
+     * @param DateTime $startTime
+     * @param DateTime $endTime
+     * @return string[][]
+     */
+    public function perArticle(
```
MaxSem: We generally use verb function names, so shouldn't this be something like getPerArticle()?

src/AppBundle/Service/EventProcessor.php
```diff
@@ -246,11 +252,13 @@ private function setContributionStats(): void
             $this->createOrUpdateEventStat('pages-created', $this->pagesCreated);
             $this->createOrUpdateEventStat('pages-improved', $this->pagesImproved);
             $this->createOrUpdateEventStat('byte-difference', $this->byteDifference);
+            $this->createOrUpdateEventStat('pages-created-pageviews', $this->pagesCreatedPageviews);
```
MaxSem: This is not really contribution stats.

samwilson: What do you mean?

src/AppBundle/Repository/EventWikiRepository.php
samwilson: Good point. Done.

src/AppBundle/Repository/PageviewsRepository.php
samwilson: Done.

src/AppBundle/Service/EventProcessor.php
MaxSem: It should be in a function other than setContributionStats()

src/AppBundle/Model/Event.php
```diff
@@ -49,6 +49,8 @@ class Event
         'byte-difference' => null,
         'pages-created' => null,
         'pages-improved' => null,
+        'pages-created-pageviews' => null,
```
MusikAnimal: This doesn't need an offset too? I think we only use them in messaging, which I know doesn't exist yet for this metric.

src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -257,6 +258,56 @@ public function getPageIds(
...
+        DateTime $start,
+        array $pageIds,
+        bool $getDailyAverage = false
+    ): int {
+        if (0 === count($pageIds)) {
+            return 0;
+        }
+        $pageviewsRepo = new PageviewsRepository();
+        $recentDayCount = Event::AVAILABLE_METRICS['pages-improved-pageviews'];
+        $thirtyDaysAgo = (new DateTime())->sub(new DateInterval('P'.$recentDayCount.'D'));
```
MusikAnimal: Since `$recentDayCount` isn't necessarily 30 days (could be changed later), I'd use a different variable name. Maybe `$offsetDate`?

src/AppBundle/Model/Event.php
samwilson: Nope, the `pages-improved-pageviews` is an average per day over (up to) the last 30 days, whereas the `pages-created-pageviews` is the cumulative total over all time since page creation till now (so there's no need for an offset).

src/AppBundle/Repository/EventWikiRepository.php
samwilson: Good point! Fixed.

PR #154, opened 2018-12-18
Split page IDs into 'created' and 'improved'

samwilson:
Bug: https://phabricator.wikimedia.org/T206817

### DIFF
--- a/src/AppBundle/Model/EventWiki.php
+++ b/src/AppBundle/Model/EventWiki.php
@@ -10,6 +10,7 @@ namespace AppBundle\Model;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Exception;
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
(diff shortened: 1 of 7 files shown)


src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -243,7 +245,14 @@ public function getPageIds(
...
 
-        $result = $this->executeQueryBuilder($rqb)->fetchAll(\PDO::FETCH_COLUMN);
+        // If only pages created or edited are being requested, limit based on the presence of a parent revision. This
+        // matches the check done in EventRepository::getEditStats().
+        if (in_array($type, ['created', 'edited'])) {
+            $typeOperator = 'edited' === $type ? '!=' : '=';
+            $rqb->andWhere("rev_parent_id $typeOperator 0");
+        }
+
+         $result = $this->executeQueryBuilder($rqb)->fetchAll(\PDO::FETCH_COLUMN);
```
samwilson: Weird, there's an extra space indent here that's not being complained about by phpcs.

samwilson:
@MusikAnimal do you have an opinion on how to fix the parameter-count thing that codeclimate is whinging about? https://codeclimate.com/github/wikimedia/eventmetrics/pull/154

MusikAnimal:
> @MusikAnimal do you have an opinion on how to fix the parameter-count thing that codeclimate is whinging about? https://codeclimate.com/github/wikimedia/eventmetrics/pull/154

We can just ignore it :) I think six parameters in this case is fine. I see it as more of a guideline than a hard rule. Would a higher limit make more sense, maybe 8? Config is at https://github.com/wikimedia/eventmetrics/blob/master/.codeclimate.yml#L5

samwilson:
@MusikAnimal can you have another look at this and see what you think?

src/AppBundle/Model/EventWiki.php
```diff
@@ -75,14 +76,21 @@ class EventWiki
...
     /**
-     * A bzcompressed string of all the page IDs. Compression is used so we don't have to store enormous strings.
-     * When pulled from the database, this may be of type resource and not string. Use self::getPages() to get an
-     * array representation of the page IDs.
+     * A bzcompressed string of all the IDs of pages created. Compression is used so we don't have to store enormous
+     * strings. When pulled from the database, this may be of type resource and not string. Use self::getPages() to get
+     * an array representation of the page IDs.
      * @see https://secure.php.net/manual/en/function.bzcompress.php
-     * @ORM\Column(name="ew_pages", type="blob", length=512, nullable=true)
+     * @ORM\Column(name="ew_pages_created", type="blob", length=512, nullable=true)
```
MusikAnimal: I don't remember why I added `length=512`. From the migration you generated it seems this has no effect, but we should still remove it for clarity. Doesn't have to be with this PR.

samwilson: Oh yes, that's a good point. I'll remove it now. Blobs can't have length; maybe it was to be used for an index length? But we're not indexing these so it doesn't matter.

PR #155, opened 2018-12-19
Byte difference stat

MaxSem:
Bug: https://phabricator.wikimedia.org/T206820

### DIFF
--- a/src/AppBundle/Model/Event.php
+++ b/src/AppBundle/Model/Event.php
@@ -46,6 +46,7 @@ class Event
         'new-editors' => 15,
         'retention' => 7,
         'edits' => null,
+        'byte-difference' => null,
         'pages-created' => null,
         'pages-improved' => null,
         'files-uploaded' => null,
@@ -59,7 +60,7 @@ class Event
      */
     public const WIKI_FAMILY_METRIC_MAP = [
         '*' => ['edits', 'new-editors', 'retention'],
-        'wikipedia' => ['pages-created', 'pages-improved'],
+        'wikipedia' => ['pages-created', 'pages-improved', 'byte-difference'],
         'commons' => ['files-uploaded', 'file-usage'],
         'wikidata' => ['items-created', 'items-improved'],
     ];
--- a/src/AppBundle/Repository/EventRepository.php
+++ b/src/AppBundle/Repository/EventRepository.php
@@ -395,4 +395,24 @@ class EventRepository extends Repository
 
         return ltrim(implode(',', $usernames), ',');
     }
+
+    /**
+     * Get the status of the existing job for this event, if any.
+     *
+     * @param Event $event
+     * @return bool|null true if job has been started, false if queued, null if nonexistent.
+     */
+    public function getJobStatus(Event $event): ?bool
+    {
+        $conn = $this->getEventMetricsConnection();
+        $rqb = $conn->createQueryBuilder();
+        $eventId = $event->getId();
+
+        $rqb->select('job_started')
+            ->from('job')
+            ->where("job_event_id = $eventId");
+
+        $ret = $this->executeQueryBuilder($rqb, -1)->fetch();
+        return isset($ret['job_started']) ? (bool)$ret['job_started'] : null;
+    }
 }
--- a/src/AppBundle/Repository/Repository.php
+++ b/src/AppBundle/Repository/Repository.php
@@ -85,6 +85,8 @@ abstract class Repository extends EntityRepository
     public function setContainer(Container $container): void
     {
         $this->container = $container;
+        // HACK: DI should do this automatically via setCachePool() but it doesn't
+        $this->cache = $container->get('cache.app');
     }
 
     /**
@@ -327,10 +329,10 @@ abstract class Repository extends EntityRepository
      */
     protected function getTableName(string $name, ?string $suffix = null): string
     {
-        if (null !== $suffix) {
-            return $name.'_'.$suffix;
-        } elseif ('' === $suffix) {
+        if ('' === $suffix) {
             return $name;
+        } elseif (null !== $suffix) {
+            return $name.'_'.$suffix;
         }
 
         // For 'revision' and 'logging' tables (actually views) on the WMF replicas,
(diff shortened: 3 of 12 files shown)


src/AppBundle/Repository/EventRepository.php
```diff
@@ -396,6 +396,135 @@ private function getUsernamesSql(Event $event): string
...
+        return $result;
+    }
+
+    /**
+     * Returns an SQL-friendly stringified list of identifiers
+     *
+     * @param int[] $ids
+     * @return string
+     */
+    private function idList(array $ids): string
```
MusikAnimal: This is nice! I would even put it in `Repository`, as I think we could it use it in other places

src/AppBundle/Repository/EventRepository.php
```diff
@@ -396,6 +396,135 @@ private function getUsernamesSql(Event $event): string
...
+    }
+
+    /**
+     * @param Event $event
+     * @return mixed[]
+     */
+    private function getFirstLastRevisions(Event $event): array
+    {
+        /** @var string[][] $revs */
+        $revs = $this->getRevisions($event);
```
MusikAnimal: I imagine going through every edit is somewhat slow... I wonder if there's a way to get the length of the first/last revision using dedicated SQL queries? We have the page IDs on hand, so you could loop through those and make queries accordingly.

MaxSem: I was assuming this was going to be the intended use case for revision IDs that we store.

src/AppBundle/Repository/EventRepository.php
MaxSem: Done.

src/AppBundle/Repository/EventRepository.php
MusikAnimal: I'll reply here too... I don't think the plan is to store revision IDs, if that's what you mean? Only page IDs. So I envisioned using one or two queries to fetch `rev_len` from the first/last revision with `rev_page IN (...)`.

src/AppBundle/Model/Event.php
```diff
@@ -78,6 +79,8 @@ class Event
         'file-usage',
         'items-created',
         'items-improved',
+        // TODO: remove this
```
aezell: Why is it a `TODO` to get rid of this?

src/AppBundle/Repository/EventRepository.php
```diff
@@ -396,6 +396,83 @@ private function getUsernamesSql(Event $event): string
...
+              AND cur.rev_user_text IN (:usernames)
+            ORDER BY cur.rev_timestamp ASC
+            LIMIT 1";
+
+        $outerSql = "SELECT SUM(after) - SUM(before_)
+            FROM (
+                SELECT ($after) after, ($before) before_
+                    FROM $dbName.$pageTable
+                    WHERE page_id IN (:pageIds)
+                ) t1";
```
aezell: Did you try using the page IDs in the `after` and `before` queries and then just do the calculation in the PHP code rather than in the SQL? I'm wondering if the query time is equivalent between the two approaches. I also wonder if that approach wouldn't make the code a bit simpler.

src/AppBundle/Model/Event.php
MaxSem: This is list of metrics that will be shown in UI. Displaying them is helpful when debugging, but not needed in production.

src/AppBundle/Repository/EventRepository.php
MaxSem: You mean return all the revisions during the event?

MaxSem:
This is ready for review.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -396,6 +397,83 @@ private function getUsernamesSql(Event $event): string
...
+     * @param string $dbName
+     * @param int[] $pageIds
+     * @return int
+     */
+    private function getBytesInternal(Event $event, string $dbName, array $pageIds): int
+    {
+        $revisionTable = $this->getTableName('revision');
+        $pageTable = $this->getTableName('page');
+        $userIds = $event->getParticipantIds();
+        $usernames = array_column($this->getUsernamesFromIds($userIds), 'user_name');
```
MusikAnimal: This is already fetched and cached in EventProcessor. So you might pass it in to `getBytesChanged`. We should probably add query caching to Repository::getUsernamesFromIds() too, or instead... not sure what the ideal solution is.

src/AppBundle/Repository/Repository.php
```diff
@@ -327,10 +329,10 @@ public function getUsernameFromId(int $userId): ?string
      */
     protected function getTableName(string $name, ?string $suffix = null): string
     {
-        if (null !== $suffix) {
-            return $name.'_'.$suffix;
-        } elseif ('' === $suffix) {
+        if ('' === $suffix) {
             return $name;
+        } elseif (null !== $suffix) {
+            return $name.'_'.$suffix;
```
MusikAnimal: Good catch, thank you :)

src/AppBundle/Repository/EventRepository.php
MaxSem: That would require moving the byte difference stuff to `setContributionsWikipedias` which already has a scary signature - do you think it's worth it?

MaxSem: Meh, turned the parameters into fields.

MusikAnimal:
Code looks amazing! I really like making the stats in EventProcessor into properties. Makes mucho sense.

I see one problem -- apparently we aren't counting bytes added in the first revision. For instance, an enwiki event that starts at 2018-06-09 05:00 (America/New York) and ends at 2018-06-12 04:59, with MusikAnimal as the only participant, and limited to the category "Parks in Brooklyn". This is similar to the event used in the test case.

If you review the edits: https://en.wikipedia.org/w/index.php?title=Domino_Park&action=history&offset=20180610150000 and summate it manually, you end up with 4641 bytes. Yet in the database I see Event Metrics recorded 4206, which is 4641 - 435 (initial revision) = 4206. Basically in the "before" query we need to get the rev_len of the revision _before_ the first one: https://github.com/wikimedia/eventmetrics/pull/155/files#diff-6290201f755154950df266ebb23cdd65R341 Of course, in this case there is no prior revision, so the value should be 0. I don't immediately have any good ideas of how to do this... :/

MaxSem:
Turning it into LEFT JOIN did the trick, thanks for testing the heck out of it!

MusikAnimal:
As you would say: _weeeeeeeeeeeeeeee!!_

PR #159, opened 2019-01-09
Prevent empty page IDs when decompressing bzcompressed data

samwilson:
When we switched to two page ID columns the 2nd one is initally
populated with NULLs. When this is retrieved and decompressed
it results in an empty string, which was being passed through
in the pageIds arrays and breaking SQL.

This patch fixes it by removing any empty elements in the returned
array.

Bug: https://phabricator.wikimedia.org/T213230

### DIFF
--- a/src/AppBundle/Model/EventWiki.php
+++ b/src/AppBundle/Model/EventWiki.php
@@ -321,7 +321,8 @@ class EventWiki
             ? stream_get_contents($this->$propertyName)
             : $this->$propertyName;
 
-        return explode(',', bzdecompress($blob));
+        // Decompressing the empty string or null results in an empty string, so array_filter removes this.
+        return array_filter(explode(',', bzdecompress($blob)));
     }
 
     /**
@@ -341,6 +342,13 @@ class EventWiki
             $this->$propertyName = null;
             return;
         }
-        $this->$propertyName = bzcompress(implode(',', $ids));
+        $filteredIds = array_filter($ids, function ($id) {
+            if (!is_numeric($id)) {
+                // Will be filtered out.
+                return null;
+            }
+            return (int)$id;
+        });
+        $this->$propertyName = bzcompress(implode(',', $filteredIds));
     }
 }
--- a/tests/AppBundle/Model/EventWikiTest.php
+++ b/tests/AppBundle/Model/EventWikiTest.php
@@ -147,5 +147,11 @@ class EventWikiTest extends EventMetricsTestCase
         $wiki->setPagesEdited([4, 5, 6]);
         static::assertEquals([4, 5, 6], $wiki->getPagesEdited());
         static::assertEquals([1, 2, 3, 4, 5, 6], $wiki->getPages());
+
+        // Make sure that empty strings are removed, and strings are cast to integers.
+        $wiki->setPagesCreated(['']);
+        static::assertEquals([], $wiki->getPagesCreated());
+        $wiki->setPagesCreated(['', '123', null, '', 456, 'foo']);
+        static::assertEquals([123, 456], $wiki->getPagesCreated());
     }
 }


src/AppBundle/Model/EventWiki.php
```diff
@@ -321,7 +321,8 @@ protected function getPageIds(string $type): array
             ? stream_get_contents($this->$propertyName)
             : $this->$propertyName;
 
-        return explode(',', bzdecompress($blob));
+        // Decompressing the empty string or null results in an empty string, so array_filter removes this.
+        return array_filter(explode(',', bzdecompress($blob)));
```
mooeypoo: This is a good sanity check, but I'm wondering if we should also remove NULL and empty values when **storing** the data; we've gone to some length to minimize the stored string so to accommodate large ID pools -- it sounds like it could be a good idea to prevent "waste" of space in the storage too, to begin with?

tests/AppBundle/Model/EventWikiTest.php
```diff
@@ -147,5 +147,9 @@ public function testPages(): void
         $wiki->setPagesEdited([4, 5, 6]);
         static::assertEquals([4, 5, 6], $wiki->getPagesEdited());
         static::assertEquals([1, 2, 3, 4, 5, 6], $wiki->getPages());
+
+        // Make sure that empty strings are removed.
+        $wiki->setPagesCreated(['']);
```
mooeypoo: I think this test would be more representative if you added empties in between numbers, as it will usually come from the database?

Also, are we really expecting numbers, or strings?

src/AppBundle/Model/EventWiki.php
samwilson: No, I'm pretty sure we're never actually storing empty IDs, at least I haven't been able to make it do that. We're *retrieving* an empty value — a single one — only when pulling from a NULL DB value (because `bzdecompress(null)` and `bzdecompress('')` both give us an empty string). Which means the `array_filter` is slightly unnecessary (we could instead check for empty string, but this is smaller; I don't know if there's performance differences; should I switch it?).

tests/AppBundle/Model/EventWikiTest.php
samwilson: Yeah, good idea. I think we never do save anything other than arrays of integers, but "be liberal in what you accept" and all that...

I'll update it.

src/AppBundle/Model/EventWiki.php
mooeypoo: Then why was the bug with (,xxx,yyy) ? That suggested that you have an array of [ null, xxx, yyy ] not just one null, no?

mooeypoo:
This looks good to me.

I'm still a little confused about how we store those IDs, and whether we might be storing nulls or empty values. We clear them out on 'get' so it shouldn't really matter anymore ,but we might want to verify that we're not storing unnecessary empty or null values in the DB.

That can be a general cleanup followup task, though.

mooeypoo:
The travis failure looks like it has to do with regenerating assets? It's a little weird given you didn't change anything related to assets, but maybe rebase-based?

src/AppBundle/Model/EventWiki.php
MusikAnimal: Before there was only one field, `page_ids`, so `explode(',', bzdecompress(null))` gives an array with one blank value. Then in EventRepository we use `implode(',', $pageIds)` for use in SQL, whilst [ignoring empty strings](https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/Repository/EventRepository.php#L352) (meaning there were no page IDs). With #154, however, we have two page ID fields. `EventWiki::getPages()` merges the arrays and you can end up with `['', 1, 2]`. The `implode` function will produce `,1,2`, hence the broken SQL. So we're not storing blank values for page IDs as with `[null, 1, 2]`, rather we're correctly storing `null` in one field and `bzcompress(implode(',', [1, 2]))` in the other.

MusikAnimal:
> The travis failure looks like it has to do with regenerating assets? It's a little weird given you didn't change anything related to assets, but maybe rebase-based?

I think in an earlier commit we pushed asset changes that were generated using an older version of Node. Travis is on 11.4.0 https://github.com/wikimedia/eventmetrics/blob/master/.travis.yml#L11

@samwilson ran into this issue with SVG Translate I think; the solution being to enforce a Node version in development via nvm?

At any rate, if you rebase in master tests will now pass, following ef7897e62220ee0b123eaa9beee91e68c0ba8607

samwilson:
Yay, it's passing now.

PR #160, opened 2019-01-09
Move nvm version number to .nvmrc

samwilson:
This makes it easier for developers to switch to the required
Node version by just typing 'nvm use'.


MusikAnimal:
I'm very confused about that test failure!

samwilson:
I think that's because there was some weird short term connectivity problem with the replica DB.

PR #161, opened 2019-01-10
Store the status of Jobs and report failures

MusikAnimal:
Change the job_started (bool) column into a tinyint that represents the
status of the Job. We do this so that we can flag it as failed.

This is basically a regression fix. Before we deployed the 'proper' job
queue system, a failed job would be reported in the repose of the
request to start the job. Now it is a true background process so we must
store the job state in order to determine if it failed.


MusikAnimal:
FYI this is a subset of #153 that doesn't include design changes.

src/AppBundle/Model/Event.php
```diff
@@ -752,6 +752,15 @@ public function addJob(Job $job): void
         $this->jobs->add($job);
     }
 
+    /**
+     * A convenience method to get the first Job (which should be the only one, for now).
+     * @return Job|false
+     */
+    public function getJob()
+    {
+        return $this->jobs->first();
```
aezell: I wonder if we should have some deterministic ordering here to future proof against bugs with multiple jobs.

aezell: I think there's also a place where this could be used on line 189 of EventDataController.php.

MusikAnimal: Yeah, so I think we talked about this some; I wanted to change it so that it's a one-to-one relationship, so that there can ONLY be one job. That's a big-ish change, and I worry if there will be some point where we do want multiple jobs. So this is sort of like a interim thing until we decide on it once and for all.

MusikAnimal: I just got tired of writing $event->getJobs()->first(), is all :)

PR #162, opened 2019-01-10
Fix phpcs failures

MaxSem:


MusikAnimal:
Thanks, I put this in #161 too. Apparently `composer lint` doesn't cause the CI build to fail. I'll work on that now.

PR #165, opened 2019-01-14
Add code to track usage to piwik.wikimedia.org

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T213735


MusikAnimal:
I don't think we'll really know if this works until it's on production. I think it is looking only for requests from eventmetrics.wmflabs.org. I'm using the exact code Nuria gave me, so I've confidence :)

aezell:
Can you claim this task and move it into the right status?

https://phabricator.wikimedia.org/T213735

PR #167, opened 2019-01-16
Update to UI feedback for when a job fails

MusikAnimal:
Delete job after the modal is shown, so they aren't presented with it
everytime they go back to the page.

Fix exception handling in JobHandler.

Specify HTTP methods on API endpoints.

Various code cleanup in EventDataController.

Bug: https://phabricator.wikimedia.org/T207776


MusikAnimal:
By the way, you can test the "timed out" error by setting `app.query_timeout: 1` in your parameters.yml. That should show an error on any medium-sized event. For the "uknown" error, you can put syntax errors in EventProcessor, for instance.

app/Resources/assets/js/eventshow.js
```diff
@@ -88,6 +88,17 @@ eventmetrics.eventshow.setState = function (state) {
     } else {
         $('body').removeClass('disabled-state');
     }
+
+    // Show the modal if needed.
```
samwilson: This seems to still only be called for *some* 500 errors for the job-status request. Should `.fail()` also be called for line 113 below?

MusikAnimal: The job status endpoint will return a 200 regardless of the status of the job, since the request to get the status itself was successful. https://github.com/wikimedia/eventmetrics/pull/167/files#diff-a03fda7eed4e489eca7b27603175f4e8L209

That said, it's possible the endpoint will error out for other reasons. Though if it does I'm not sure what we'd do :/ This endpoint is necessary so that the client updates the view accordingly. If it fails it's just going to sit there indefinitely with no UI feedback. I think the safest thing is probably to just let it respond only to successful requests as it does now (say one request fails, but the next one succeeds). We'll get an email if something went wildly wrong on the backend.

app/Resources/assets/js/eventshow.js
```diff
@@ -88,6 +88,17 @@ eventmetrics.eventshow.setState = function (state) {
     } else {
```
mooeypoo: Not really a blocker or anything, but you can also use $( 'body' ).toggleClass( 'disabled-state', 'started' === state );

app/Resources/assets/js/eventshow.js
samwilson: > If it fails it's just going to sit there indefinitely with no UI feedback.

Isn't that what this new alert is fixing though?

I was just thinking that below in `pollFunc` it could also set a failure state, e.g.:

        .fail(function () {
            eventmetrics.eventshow.setState('failed-unknown');
        })

But yeah, it wouldn't really help for whatever failure that'd be producing.

I'm happy for this to be merged as-is.

MusikAnimal: Ah, I see. That does make sense... but I'm theorizing some scary scenarios:
* `/events/job-status/` for some reason fails
* They are shown the "failed-unknown" modal, which says to try again
* They try again, which deletes the old job (but not the process! yet...) so now you have two sets of long-running queries

All the while the would-be subsequent request to /events/job-status/ could be successful. If we did a retry count I think showing failed-unknown would be fine, but that's maybe too much for what should be an unlikely scenario.

app/Resources/assets/js/eventshow.js
MusikAnimal: I like it! I'll make this change. I need to rebase in master anyway.

app/Resources/assets/js/eventshow.js
samwilson: Oh yes, that makes much more sense.

PR #170, opened 2019-01-18
Fix double logging; no emails in dev environment

MusikAnimal:
We had monolog configuration in both config.yml and the environment-
specific config (e.g. config_dev.yml). This made everything be logged
twice. This commit makes each environment config have its own monolog
configuration, which has the side effect of ensuring no emails are sent
in the dev environment.


samwilson:
Thank you!

And I'm sorry... I think I introduced this duplication? Thanks for fixing. :)

PR #171, opened 2019-01-22
Include local file uploads in addition to Commons

MusikAnimal:
This extends the current files-uploaded and file-usage metrics to be
computed on all content wikis and not just Commons.
    
Other very minor code and comment nitpicks.

Bug: https://phabricator.wikimedia.org/T206819


MusikAnimal:
Turns out we're not using page IDs to show files in the Edit List, instead putting `page_namespace = 6 and rev_parent_id = 0` in the query. That makes this a lot easier. It probably should use page IDs, especially as we want perma-stats after the event has ended (say some files got deleted), but save that for a separate PR.

PR #172, opened 2019-01-22
Prevent forms from being submitted more than once

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T207224


aezell:
Would it be better to disable the submit button and change its style in addition to catching subsequent clicks? That way, we are letting the user know that it worked instead of them clicking furiously.

MusikAnimal:
> Would it be better to disable the submit button and change its style in addition to catching subsequent clicks? That way, we are letting the user know that it worked instead of them clicking furiously.

I don't see any reason not to! Done.

PR #173, opened 2019-01-23
Participant counts from page IDs

MaxSem:
Bug: https://phabricator.wikimedia.org/T208546

### DIFF
--- a/src/AppBundle/Repository/EventWikiRepository.php
+++ b/src/AppBundle/Repository/EventWikiRepository.php
@@ -427,4 +427,28 @@ class EventWikiRepository extends Repository
 
         return (int)$res->fetchColumn();
     }
+
+    /**
+     * Get the list of users participating in an event with no predefined user list
+     *
+     * @param string $dbName
+     * @param int[] $pageIds
+     * @param Event $event
+     * @return string[]
+     */
+    public function getUsersFromPageIDs(string $dbName, array $pageIds, Event $event): array
+    {
+        $revisionTable = $this->getTableName('revision');
+        $rqb = $this->getReplicaConnection()->createQueryBuilder();
+        $rqb->select('DISTINCT(rev_user_text)')
+            ->from("$dbName.$revisionTable")
+            ->where('rev_page IN (:pageIds)')
+            ->andWhere('rev_user <> 0')
+            ->andWhere('rev_timestamp BETWEEN :start AND :end')
+            ->setParameter('pageIds', $pageIds, Connection::PARAM_INT_ARRAY)
+            ->setParameter('start', $event->getStart()->format('YmdHis'))
+            ->setParameter('end', $event->getEnd()->format('YmdHis'));
+
+        return $this->executeQueryBuilder($rqb)->fetchAll(\PDO::FETCH_COLUMN);
+    }
 }
--- a/src/AppBundle/Service/EventProcessor.php
+++ b/src/AppBundle/Service/EventProcessor.php
@@ -45,6 +45,9 @@ class EventProcessor
     /** @var string[]|null Usernames of the new editors. */
     private $newEditors;
 
+    /** @var string[] Array with usernames of editors from page IDs as keys */
+    private $implicitEditors = [];
+
     /** @var string[] Unique wikis where participants have made edits. */
     protected $commonWikis;
 
@@ -93,7 +96,6 @@ class EventProcessor
         $this->log('Processing event '.$event->getId());
 
         // Generate and persist each type of EventStat/EventWikiStat.
-        $this->setNewEditors();
 
         // If there is an associated EventWiki that represents a family, we need to first find
         // the common wikis where participants have edited. From there we will create new EventWikis
@@ -107,6 +109,10 @@ class EventProcessor
 
         // Remove EventWikis that are part of a family where there are no statistics.
         $this->removeEventWikisWithNoStats();
+
+        $this->setParticipants();
+
+        $this->setNewEditors();
 
         $this->setRetention();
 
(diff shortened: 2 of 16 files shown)


src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -370,4 +370,28 @@ public function getBytesChanged(Event $event, string $dbName, array $pageIds, ar
...
+     * @return string[]
+     */
+    public function getUsersFromPageIDs(string $dbName, array $pageIds, Event $event): array
+    {
+        $revisionTable = $this->getTableName('revision');
+        $rqb = $this->getReplicaConnection()->createQueryBuilder();
+        $rqb->select('DISTINCT(rev_user_text)')
+            ->from("$dbName.$revisionTable")
+            ->where('rev_page IN (:pageIds)')
+            ->andWhere('rev_user <> 0')
```
MusikAnimal: I didn't even know about the `<>` operator! I always used `!=`. Anyway, a tiny nitpick -- it may be a little misleading here because accounts will always have IDs greater than zero, and we don't want NULL values (though I realize none exist).

src/AppBundle/Service/EventProcessor.php
```diff
@@ -352,6 +359,20 @@ private function setItemsCreatedOrImprovedOnWikidata(EventWiki $wiki, EventWikiR
         $this->createOrUpdateEventStat('items-improved', $ret['edited']);
     }
 
+    private function setUserCounts(EventWiki $wiki, EventWikiRepository $ewRepo): void
```
MusikAnimal: Not sure why phpcs didn't complain, but this method needs a docblock :)

MusikAnimal:
This seems like the right approach to me! Performance may become a problem, but we'll worry about that later :)

src/AppBundle/Service/EventProcessor.php
MusikAnimal: Ah! Are we expected to document private methods too? It feels right to me to document everything, but I get that if this were a library, private methods need not be explained because they shouldn't/can't be used.

MaxSem: Done.

src/AppBundle/Service/EventProcessor.php
```diff
@@ -399,6 +420,26 @@ private function setItemsCreatedOrImprovedOnWikidata(EventWiki $wiki, EventWikiR
...
+     * @param EventWikiRepository $ewRepo
+     */
+    private function setUserCounts(EventWiki $wiki, EventWikiRepository $ewRepo): void
+    {
+        $dbName = $ewRepo->getDbNameFromDomain($wiki->getDomain());
+
+        $pageIds = $wiki->getPages();
+        $usernames = $this->getParticipantNames();
+        if ($pageIds) {
+            if (!$usernames && $this->event->getNumCategories() > 0) {
```
MusikAnimal: lonely if? You could just use `if ($pageIds && !$usernames && $this->event->getNumCategories() > 0)`, no?

MusikAnimal:
Looks awesome! So, I'm guessing this is a stepping stone to supporting work lists?

From the task:
> Both the "Participants" and "New Editors" metrics remain fixed once the event period ends.

In our case this basically means don't run redundant queries after the event has concluded, and it doesn't look like we've done that here. To be fair, I don't think we've done this for any of the metrics that remain static ;) But here it seems rather straightforward -- only call `setParticipants()` in EventProcessor if it's before the end date or if the current participant count is zero. Did you want to do that as part of this PR?

Another thing that possibly should be done in a separate PR: making use of the new `participants` EventStat in place of where we are counting `Participant`s. That is a bit involved... you'll need to update (rough outline, may be incomplete):
* `Event::VISIBLE_METRICS`
* `Event::getNumParticipants()` (returning the value of the relevant EventStat, or `$this->participants->count()` if it does not exist)
* `ProgramRepository::getNumParticipants()`

and these views to remove participant code (since it will automatically show):
* `events/_stats_summary.html.twig`
* `events/event_summary.wikitext.twig`
* `events/show.html.twig`
* `events/_participants.html.twig`

On that note, for when manually entering participants via the form, we probably will want to set the EventStat for `participants`. Otherwise the view may say "0 participants" when they have in fact added participants, they just haven't generated the stats yet. Not sure if that makes sense... either way it's a bit of work, so in my opinion it's OK to do this in a separate PR.

MusikAnimal:
Also, ignore codeclimate. EventProcessor is going to end up having even more methods. Feel free to change or remove the `method-count` option in `.codeclimate.yml`

src/AppBundle/Service/EventProcessor.php
MaxSem: Done.

aezell:
@MaxSem Could you address this from Leon's comment: 

> only call `setParticipants()` in EventProcessor if it's before the end date or if the current participant count is zero. Did you want to do that as part of this PR?

I think that's a good idea in this PR. If you add that, I'll approve the review and we can merge.

MaxSem:
I'm not sure that this solution is going to work - if the number was calculated before the end of the event, it will save the value and not permit recalculations, even though the event continued after that and more editors have contributed.

MusikAnimal:
> I'm not sure that this solution is going to work - if the number was calculated before the end of the event, it will save the value and not permit recalculations, even though the event continued after that and more editors have contributed.

Indeed, so the logic should be to skip `setParticipants()` if `Event::getUpdated()` is after `Event::getEnd()`. Totally okay with doing this in a different PR, we just don't want to forget about it. Maybe we could combine that work with https://phabricator.wikimedia.org/T210682

MusikAnimal:
> Indeed, so the logic should be to skip `setParticipants()` if `Event::getUpdated()` is after `Event::getEnd()`. Totally okay with doing this in a different PR, we just don't want to forget about it. Maybe we could combine that work with https://phabricator.wikimedia.org/T210682

Done. I included this metric as part of T210682. It may end up being put in a different task, but I don't think we're going to forget about it. So, I'm going to merge this PR. I take it there is no QA involved whatsoever, since there's no way to trigger this logic within the application (until we add the worklist)? Just make sure you say this in the task.

PR #176, opened 2019-02-05
Fix message for pageviews of pages improved, should be avg daily

MusikAnimal:
Rename pages-improved-pageviews metric to pages-improved-pageviews-avg.

Remove incorrect translations as a result of the incorrect English msg.

Minor refactoring in EventDataController.

Fix outdated comment for EventRepository::getFileUsage()

Bug: T206692


i18n/en.json
```diff
@@ -113,7 +113,7 @@
 	"pages-created-pageviews": "Views to pages created",
 	"pages-improved": "Pages improved",
 	"pages-improved-desc": "A content page is a page in either the main or file namespace.",
-	"pages-improved-pageviews": "Views to pages improved",
+	"pages-improved-pageviews-avg": "Avg. daily views to pages improved",
```
samwilson: I don't think this message name should have the `-avg` appended.

MusikAnimal: Well, the existing message would imply the metric is total pageviews, and not an average. TWN will surely mark it as fuzzy but I figure it's better to create a new message altogether, since the meaning is different. But also I wonder if we wanted to go back to "Views to pages improved" (not an average), we'll have the same problem. I always thought it was peculiar to have a different pageviews metric for pages improved and created, and wanted to leave the possibility that this will be changed. Maybe `pages-created-pageviews-avg` would be introduced, etc.

samwilson: I totally understand changing the message name to make it more clear what it refrers to (it should have included `avg` when it was created, really). But where is this new message being used? I thought the stats messages had to match the names of the stats (although I guess that's only for the web display?). I reckon it should either be changed everywhere (i.e. change the stat name as well) or not worry about it and just add the 'avg.' info in the message text.

Also, you forgot to update qqq.json.

MusikAnimal: Ah, right... duh... yes the need message needs to match the `es_metric` column! I have changed it everywhere to use `-avg`, to leave room for the non-average metric should we want it later. This does mean we'll have some `pages-improved-pagesviews` EventStats and EventWikiStats lingering in the database, but this doesn't seem to cause problems since we always go off of Event::AVAILABLE_METRICS in some way or another. We can run a one-off query to fix the records with the old metric name, if we want.

MusikAnimal:
@samwilson Any chance you could give this another look-see? :)

PR #178, opened 2019-02-06
Bump PHPUnit (to 7.5.3) and Symfony (to 4.2)

MaxSem:


MusikAnimal:
I'm mostly OK with this, but note it does introduce several deprecations. Namely, our abstract `Controller` needs to extend `AbstractController` instead of `SymfonyController`. I tried to do this a while back and it seemed a bit involved. I think @samwilson just did the same thing for SVG Translate.

Also apparently `app/Resources/TwigBundle/views/Exception/error.html.twig` needs to be moved to something like `templates/bundles/TwigBundle/Exception/error.html.twig`. I think that's Symfony telling us it's time to adapt to the Symfony 4 directory structure.

Did you want to do any of that as part of this PR? Or is there a more urgent reason to upgrade Symfony? Or we can quietly ignore the deprecation warnings :)

aezell:
If these upgrades are going to cause deprecation warnings, we should create a task that includes this PR along with the work to remedy the deprecations. We should estimate that task and then prioritize it. Then we can have both PRs be part of that task.

MusikAnimal:
I'm going to merge as-is, and we'll handle the deprecations with another PR. 4.1 is officially unsupported now, and there are some 4.2-only features I think we want to take advantage of in some upcoming work.

PR #179, opened 2019-02-07
Set date formats to include TZ, for Moment.js

samwilson:
This explicitly adds the timezone to the event form's start and
end date fields, so that it can be correctly read by Moment.js
which otherwise treats the times as being in the browser's
timezone.

There seems to be some discrepencies with how ISO8601 formats
are handled (e.g. the 'T' separating the date and time should be
fine, but it seems to break things).

Bug: T215327


aezell:
Looks like the Travis build failed but the error is unhelpful. How do we trigger a new build?

MaxSem:
Nah, still fails.

samwilson:
I forgot to update the format in the test.

MusikAnimal:
Thanks! Tested working on my end.

PR #182, opened 2019-02-12
Count pages with uploaded files

MaxSem:
Bug: https://phabricator.wikimedia.org/T215356

### DIFF
--- a/src/AppBundle/Service/EventProcessor.php
+++ b/src/AppBundle/Service/EventProcessor.php
@@ -69,6 +69,9 @@ class EventProcessor
     /** @var int Number of pages that use the uploaded files. */
     private $fileUsage = 0;
 
+    /** @var int Pages using uploaded files. */
+    private $pagesUsingFiles = 0;
+
     /**
      * Constructor for the EventProcessor.
      * @param LoggerInterface $logger
@@ -323,7 +326,8 @@ class EventProcessor
         $this->log(">> <info>Pages improved: {$this->pagesImproved}</info>");
         $this->log(">> <info>Bytes added: {$this->byteDifference}</info>");
         $this->log(">> <info>Files uploaded: {$this->filesUploaded}</info>");
-        $this->log(">> <info>File usage: {$this->fileUsage}</info>");
+        $this->log(">> <info>Files used: {$this->fileUsage}</info>");
+        $this->log(">> <info>Pages using uploaded files: {$this->pagesUsingFiles}</info>");
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/AppBundle/Service/EventProcessor.php
```diff
@@ -380,9 +384,13 @@ private function setFilesUploaded(EventWiki $wiki, EventWikiRepository $ewRepo):
         $this->createOrUpdateEventWikiStat($wiki, 'files-uploaded', $ret);
         $this->filesUploaded += $ret;
 
-        $ret = $this->eventRepo->getFileUsage($dbName, $start, $end, $this->getParticipantNames());
+        $ret = $this->eventRepo->getUsedFiles($dbName, $start, $end, $this->getParticipantNames());
         $this->createOrUpdateEventWikiStat($wiki, 'file-usage', $ret);
         $this->fileUsage += $ret;
+
+        $ret = 0;//$this->eventRepo->getPagesUsingFiles($dbName, $start, $end, $this->getParticipantNames());
```
MusikAnimal: Lingering test code?

tests/AppBundle/Repository/EventRepositoryTest.php
```diff
@@ -0,0 +1,25 @@
...
+        $kernel = static::bootKernel();
+        /** @var \Doctrine\ORM\EntityManager $entityManager */
+        $entityManager = $kernel->getContainer()->get('doctrine')->getManager();
+        $repo = new EventRepository($entityManager);
+        $repo->setContainer($kernel->getContainer());
+
+        static::assertGreaterThan(0, $repo->getPagesUsingFile('commonswiki_p', 'Ultrasonic_humidifier.jpg'));
+        static::assertGreaterThan(0, $repo->getPagesUsingFile('enwiki_p', '2-cube.png'));
+    }
+}
```
MusikAnimal: We don't usually test repositories (comment for the EventRepository class is even marked with `@codeCoverageIgnore`), but I'm certainly not opposed to it :smiley: I think it'd be good to have some integration testing. The "extended" fixture includes local file uploads that should always have more than zero pages using the files: https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/DataFixtures/ORM/extended.yml#L47-L51 So in ProcessEventCommandTest you should be able to add a `pagesWithFilesSpec()`, like the others, making assertions against the relevant EventStat.

src/AppBundle/Service/EventProcessor.php
MaxSem: Fixed.

tests/AppBundle/Command/ProcessEventCommandTest.php
```diff
@@ -315,6 +315,14 @@ private function fileUsageSpec(): void
...
             ]);
         static::assertGreaterThan(1, $eventWikiStat->getValue());
+
+        $eventWikiStat = $this->entityManager
+            ->getRepository('Model:EventWikiStat')
+            ->findOneBy([
+                'wiki' => $this->event->getWikiByDomain('commons.wikimedia'),
+                'metric' => 'pages-using-files',
+            ]);
+        static::assertGreaterThan(0, $eventWikiStat->getValue());
```
MusikAnimal: Can we add one for enwiki too? Just for sanity :)

PR #183, opened 2019-02-13
Change the pages-improved stat to not count pages created

samwilson:
Bug: T206817

### DIFF
--- a/src/AppBundle/Service/EventProcessor.php
+++ b/src/AppBundle/Service/EventProcessor.php
@@ -348,18 +348,20 @@ class EventProcessor
         $wiki->setPagesEdited($pageIdsEdited);
 
         $pageIds = array_merge($pageIdsCreated, $pageIdsEdited);
-        $ret = $this->eventRepo->getEditStats($dbName, $pageIds, $start, $end, $usernames);
+        $totalEditCount = $this->eventRepo->getTotalEditCount($dbName, $pageIds, $start, $end, $usernames);
 
         $diff = $ewRepo->getBytesChanged($this->event, $dbName, $pageIds, $usernames);
 
-        $this->pagesCreated += $ret['created'];
-        $this->pagesImproved += $ret['edited'];
-        $this->edits += $ret['edits'];
+        $totalCreated = count($pageIdsCreated);
+        $totalEdited = count($pageIdsEdited);
+        $this->pagesCreated += $totalCreated;
+        $this->pagesImproved += $totalEdited;
+        $this->edits += $totalEditCount;
         $this->byteDifference += $diff;
 
-        $this->createOrUpdateEventWikiStat($wiki, 'edits', $ret['edits']);
-        $this->createOrUpdateEventWikiStat($wiki, 'pages-created', $ret['created']);
-        $this->createOrUpdateEventWikiStat($wiki, 'pages-improved', $ret['edited']);
+        $this->createOrUpdateEventWikiStat($wiki, 'edits', $totalEditCount);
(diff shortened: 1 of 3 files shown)


samwilson:
> You're counting revisions that didn't create a page here. Let me know if I can help figuring this out.

I'd love you to take a look at it! My brain is full of mush at the moment, with a cold.

It's meant to be counting revisions that didn't create a page, isn't it? i.e. those are edits to existing pages. Compare to https://github.com/wikimedia/eventmetrics/blob/0a8a8f3d8d012d8ff160610b3985a221b20c0ca6/src/AppBundle/Repository/EventWikiRepository.php#L259

samwilson:
> That begs the question: do we even need to fetch `edited` and `created` in EventRepository::getEditStats()? We can simply `count()` the page IDs.

Yeah, I was wondering about that. Would save the extra query. Is getEditStats needed for anything else?

MusikAnimal:
> > That begs the question: do we even need to fetch `edited` and `created` in EventRepository::getEditStats()? We can simply `count()` the page IDs.
> 
> Yeah, I was wondering about that. Would save the extra query. Is getEditStats needed for anything else?

The number of edits, that's it I think.

samwilson:
Take a look at this now, see what you think. I deleted `EventRepository::getEditStats()`.

src/AppBundle/Service/EventProcessor.php
```diff
@@ -348,18 +348,19 @@ private function setContributionsWikipedias(EventWiki $wiki, EventWikiRepository
...
-        $ret = $this->eventRepo->getEditStats($dbName, $pageIds, $start, $end, $usernames);
-
         $diff = $ewRepo->getBytesChanged($this->event, $dbName, $pageIds, $usernames);
 
-        $this->pagesCreated += $ret['created'];
-        $this->pagesImproved += $ret['edited'];
-        $this->edits += $ret['edits'];
+        $totalCreated = count($pageIdsCreated);
+        $totalEdited = count($pageIdsEdited);
+        $total = count($pageIds);
```
MusikAnimal: What about `$totalCreated + $totalEdited`, then we can remove line 350?

src/AppBundle/Service/EventProcessor.php
```diff
@@ -348,18 +348,19 @@ private function setContributionsWikipedias(EventWiki $wiki, EventWikiRepository
...
+        $this->pagesCreated += $totalCreated;
+        $this->pagesImproved += $totalEdited;
+        $this->edits += $total;
```
MusikAnimal: `pagesCreated` is the number of unique pages improved, `pagesCreated` is the number of unique pages created. Neither is a count of the edits made to those pages, right? I don't think we're querying for that anymore (it was one of the things `getEditStats()` did).

samwilson: doh! Yes, you're quite right. :) I'll fix it up.

src/AppBundle/Service/EventProcessor.php
samwilson: I think we still need $pageIds to get the bytes changed. And I've just readded getEditStats too...

samwilson:
> Did we want to make `getEditStats()` only count edits, and remove `COUNT(DISTINCT(rev_page)) AS edited` and `IFNULL(SUM(CASE WHEN rev_parent_id = 0 THEN 1 ELSE 0 END), 0) AS created`? The "edited" and "created" aren't being used anymore, so maybe even rework `getEditStats()` to be `getTotalEditCount()` and have it return an `int`. What do you think?

Good idea, I thought exactly the same thing last night after I turned the computer off. :)

Done now, see what you think.

PR #185, opened 2019-02-13
Add method to get pageviews of a single article

MusikAnimal:
This new method will be used in the Pages Created report

Bug: T205502


MusikAnimal:
Will be bundled with #187

PR #186, opened 2019-02-14
Show current branch and link to tree on local/staging within site notice

MusikAnimal:
This makes it more clear whose work you're testing.

### DIFF
--- a/src/AppBundle/Twig/AppExtension.php
+++ b/src/AppBundle/Twig/AppExtension.php
@@ -9,6 +9,7 @@ namespace AppBundle\Twig;
 
 use IntlDateFormatter;
 use NumberFormatter;
+use Symfony\Component\Process\Process;
 
 /**
  * Various Twig functions and filters.
@@ -47,6 +48,7 @@ class AppExtension extends Extension
             new \Twig_SimpleFunction('isAdmin', [$this, 'isAdmin']),
             new \Twig_SimpleFunction('shortHash', [$this, 'gitShortHash']),
             new \Twig_SimpleFunction('hash', [$this, 'gitHash']),
+            new \Twig_SimpleFunction('branch', [$this, 'gitBranch']),
             new \Twig_SimpleFunction('wikiPath', [$this, 'wikiPath']),
         ];
     }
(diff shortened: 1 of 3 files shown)


src/AppBundle/Twig/AppExtension.php
```diff
@@ -85,11 +86,20 @@ public function gitShortHash(): string
...
         return exec('git rev-parse HEAD');
     }
+
+    /**
+     * Get the currently checked-out Git branch.
+     * @return string
+     */
+    public function gitBranch(): string
+    {
+        return exec('git rev-parse --symbolic-full-name --abbrev-ref HEAD');
```
samwilson: I think this should use [Process](https://symfony.com/doc/current/components/process.html) instead.

MusikAnimal: Ah man, if we were on Symfony 4.2 we could use `Process::fromShellCommandline('...')->run()`. Instead we need to instantiate first, two liner versus one. Unless there's an easier way?

MusikAnimal: 3-liner even, but actually I think with Symfony 4.2 that would only bring it down to 2 lines (after `run()` you still need to call `->getOutput()`). No biggie

MusikAnimal:
@samwilson Changed it to use `Process`. About the bundle... would it include the siteNotice code too? Or would we just want to make a Twig helper to generate the "Git branch: master (sha)" message itself? I think that would be pretty neat!

MusikAnimal:
This looks super easy to add to the bundle, doing that now!

MusikAnimal:
This is turning into a rabbit hole. Better we at least wait until Event Metrics is on Symfony 4.2 so that we can use `Process::fromShellCommandline`. Meanwhile I feel a slight sense of urgency to get this deployed because there are so many different things being tested :/

aezell:
I agree with @MusikAnimal in that we can pull this into the bundle at a later date.

MusikAnimal:
The one code review issue was addressed and this is a low-risk, staging-only patch, so I'm going to boldly merge. Shame me accordingly :)

MusikAnimal:
The one code review issue has been addressed, and this is a low-risk, staging-only patch, so I am going to boldly merge. Shame me accordingly :)

PR #187, opened 2019-02-14
Pages created report

MusikAnimal:
Add wikitext export of Pages Created report
    
Refactor EventRepository::getPageviews() to use a helper method such
that we can get pageviews of a single article, as needed.
    
Bug: T205502


src/AppBundle/Controller/EventDataController.php
```diff
@@ -94,6 +99,29 @@ public function eventSummaryAction(): Response
         return $this->getFormattedResponse($format, 'event_summary', ['event' => $this->event]);
     }
 
+    /**
+     * Pages Created report.
+     * @Route("/programs/{programId}/events/{eventId}/pages-created", name="EventPagesCreated")
+     * @param EventWikiRepository $ewRepo
+     * @return Response
+     */
+    public function pagesCreatedAction(EventWikiRepository $ewRepo): Response
```
samwilson: It's not a huge issue given that nothing will link to it at that point, but if someone goes to `/pages-created` before an update has been done, they get a `Call to a member function setTimezone() on null` error because there's no `event_updated_at` date.

MusikAnimal: @samwilson Nice catch! Have a look now :) Before filtering and redirecting is a bit easier in Symfony 4.2! I still really miss how simple it was in Ruby on Rails... ~50 fewer lines of code, and without the need for a separate class.

Anyway I've just got it redirecting to the Event Summary page if no update has been done yet, which is what we were doing for the Edit List page (or related reports).

MusikAnimal:
@samwilson When you get a chance could you have another look at this? I "dismissed" your older review, I guess that's the workflow GitHub wants you to do? (I think it would be better if it auto-dismissed after any code changes...)

MusikAnimal:
Thanks!

PR #193, opened 2019-02-20
Rename 'Edit list' page to 'All edits'

MusikAnimal:
Bug: T215212


MaxSem:
You didn't really need to remove the message from anything other than en and qqq, TWN does that for yourself.

PR #194, opened 2019-02-20
Add new Command to process all events

MusikAnimal:
This is used when we introduce breaking changes to how stats are
generated. It should be used only when necessary.

Bug: T216640


MusikAnimal:
The command is `php bin/console app:process-all-events`, with the optional flag `--skip-spawn` if you want the cron to completely handle it, and not attempt to update events immediately.

For future reference you can pass `--help` to any command to see what options are available.

PR #195, opened 2019-02-21
Add Phan static checking

MaxSem:
The tests produce too much noise for now so not enabling them in CI,
however this commit includes a start of fixing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,16 +22,18 @@
     },
     "require": {
         "php": "^7.2",
+        "ext-ast": "*",
         "ext-PDO": "*",
         "ext-bz2": "*",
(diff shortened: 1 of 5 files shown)


MaxSem:
Done.

composer.json
```diff
@@ -22,16 +22,18 @@
     },
     "require": {
         "php": "^7.2",
+        "ext-ast": "*",
```
MusikAnimal: This should be under require-dev, no? Also need to run `composer update` to make the lock file in sync

MusikAnimal:
Phan is now in the repo, but not part of the test suite yet as there are many things still to fix

PR #198, opened 2019-02-22
Add 'Avg. daily views to pages that have uploaded files' metric

MusikAnimal:
Required some reworking of how the 'Unique pages with uploaded files'
was implemented.

Bug: T206700


MusikAnimal:
You can ignore the Scrutinizer failure. It sucks that our quality score has gone down but I'm not about to refactor EventProcessor just for this task.

PR #201, opened 2019-02-28
Fix timestamps in Pages Created report query

MusikAnimal:
Should be in UTC.

Bug: T205502


mooeypoo:
I'm a little confused about the relationship between the timestamp fix and the cache keys? The commit message is confusing, and the code includes both. Are they related? is fixing cache also fixing the timestamps?

MusikAnimal:
> I'm a little confused about the relationship between the timestamp fix and the cache keys? The commit message is confusing, and the code includes both. Are they related? is fixing cache also fixing the timestamps?

Moved to #202

I just noticed that the Pages Created report uses the page IDs in the cache key, and this could end up being too long. Two separate fixes broadly related, I guess, but indeed it's deserving of a different PR.

PR #204, opened 2019-03-01
Pages Created CSV export

MusikAnimal:
Minor refactoring in EventDataController

Bug: T206058

### DIFF
--- a/src/AppBundle/Controller/EventDataController.php
+++ b/src/AppBundle/Controller/EventDataController.php
@@ -83,16 +83,7 @@ class EventDataController extends EntityController
     public function eventSummaryReportAction(): Response
     {
         $format = 'csv' === $this->request->query->get('format') ? 'csv' : 'wikitext';
-        $response = $this->getFormattedResponse($format, 'event_summary', ['event' => $this->event]);
-        $eventName = $this->getFilenameFriendlyEventName();
-        if ('csv' === $format && '' !== $eventName) {
-            $response->headers->set(
-                'Content-Disposition',
-                "attachment; filename=\"summary_{$eventName}.csv\""
-            );
-        }
-
-        return $response;
+        return $this->getFormattedResponse($format, 'event_summary', ['event' => $this->event]);
     }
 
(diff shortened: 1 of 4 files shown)


src/AppBundle/Controller/EventDataController.php
```diff
@@ -136,6 +127,15 @@ private function getFormattedResponse(string $format, string $template, array $p
         $contentType = $formatMap[$format] ?? 'text/html';
         $response->headers->set('Content-Type', $contentType);
 
+        // Prettier file names for CSV.
+        $eventName = $this->getFilenameFriendlyEventName();
```
samwilson: I know it's not being introduced in this patch, but if an event name contains double quotes then it gets truncated to the first of em.

samwilson: (I meant to make this comment on the header-setting string construction below, actually.)

MusikAnimal: Hmm, I don't quite follow. What gets truncated? An event with the title `"Foo bar"` produces `event_summary__Foo_bar_.csv`. The double underscores aren't ideal I guess but I don't think it's a problem.

MusikAnimal:
I'll go ahead and merge with Sam's approval. If we uncover an issue with the file names we can take care of that with a different PR.

src/AppBundle/Controller/EventDataController.php
samwilson: Hm, odd. I have an event called `Test Event "One"` and the header that's sent with the CSV is `attachment; filename="pages_created_Test_Event_"One".csv"` and I'm prompted to save a file called `pages_created_Test_Event_`... maybe this is something to do with different browser's handling of the wayward quotation marks?

MusikAnimal: That is weird! And that should definitely be fixed. I've created https://github.com/wikimedia/eventmetrics/pull/205 but I can't repro so not sure if that does it.

PR #205, opened 2019-03-01
Remove double quotes from title in CSV file name

MusikAnimal:
See https://github.com/wikimedia/eventmetrics/pull/204#discussion_r261463312


MaxSem:
Actually, double quotes are reserved on Windows.

samwilson:
Oh right! I didn't know that. Are there other characters we should be avoiding?

MaxSem:
I think I included them all.

MusikAnimal:
Merging per above. Thanks for the help, [Sam & Max](https://en.wikipedia.org/wiki/Sam_%26_Max)!

PR #207, opened 2019-03-01
Use age of article when calculating average pageviews

MusikAnimal:
That is, the age or the 'offset' (currently 30 days), whatever is shorter.

Depends on: #206

Bug: T206700


MusikAnimal:
**Depends on #206**. b6b55168d1773131b4ed6f5e13d2403edc8b3889 is the relevant commit up for review.

Or, if reviewers are happy with both commits, you can just merge this PR and we'll close the other :)

PR #209, opened 2019-03-05
[WIP] Fix various bugs with pageviews; some refactoring

MusikAnimal:
Account for gaps in timeseries when computing avg pageviews.

Ensure average pageviews is in relation to how long the article existed.

Refactor some pageviews logic to PageviewsRepository. In an ideal world
where we follow best practices, PageviewsRepository wouldn't have any
logic beyond fetching the data itself. But, we're breaking the same rule
in EventWikiRepository (and the others), so this refactoring at least
moves all the core pageviews logic into the same file.

Bug: T217704


MusikAnimal:
Looks / works great!

MusikAnimal:
Merged the wrong PR, meant to do that on SVG Translate!

PR #210, opened 2019-03-05
Right-align numeric data in wikitext reports

MusikAnimal:
Some cleanup to event summary wikitext template to make it more
readable.

Bug: T205502


MaxSem:
Tests need an update.

PR #212, opened 2019-03-07
URL-encode page titles in Twig templates, and use raw in CSV export

MusikAnimal:
Bug: T215923


app/Resources/views/events/pages_created.csv.twig
```diff
@@ -1,6 +1,6 @@
 "{{ msg('title') }}","{{ msg('url') }}","{{ msg('wiki') }}","{{ msg('edits-during-event') }}","{{ msg('byte-difference-during-event') }}","{{ msg('pageviews-cumulative') }}","{{ msg('pageviews-average') }}","{{ msg('incoming-links') }}","{{ msg('more-page-metrics') }}"
 {% for page in pagesCreated %}
-"{{ page.pageTitle|replace({'_': ' '}) }}","https://{{ page.wiki }}.org/wiki/{{ page.pageTitle }}",{{ page.wiki }},{{ page.edits }},{{ page.bytes }},{{ page.pageviews }},{{ page.avgPageviews }},{{ page.links }},"https://xtools.wmflabs.org/articleinfo/{{ page.wiki }}/{{ page.pageTitle }}"
+"{{ page.pageTitle|replace({'_': ' '})|raw }}","https://{{ page.wiki }}.org/wiki/{{ page.pageTitle|e('url') }}",{{ page.wiki }},{{ page.edits }},{{ page.bytes }},{{ page.pageviews }},{{ page.avgPageviews }},{{ page.links }},"https://xtools.wmflabs.org/articleinfo/{{ page.wiki }}/{{ page.pageTitle }}"
```
samwilson: Does the pageTitle in the XTools link need to be escaped as well?

And what happens if there's a double quote in the page title?

MusikAnimal: > Does the pageTitle in the XTools link need to be escaped as well?

Yes, good catch. Since XTools uses path parameters, instead of query params, I think we could even use `|raw` to make it look nice, but I've gone with `|e('url')` to be safe.

> And what happens if there's a double quote in the page title?

I tested it by just hard-coding it, replacing `page.pageTitle` with `'"foo'` or what have you. Luckily spreadsheet software doesn't seem to mind any number of double quotes, so long as the value is wrapped in double quotes.

PR #215, opened 2019-03-08
Add 3 reports to dropdown and create format dialog box

samwilson:
This adds the 'download reports' menu to the main event page (it
was already on the edit list page), and refactors it to list the
three available reports. When a menu item is clicked a modal
dialog box opens for the user to choose between CSV and wikitext
formats. Menu items are disabled if the event data is not
available yet (i.e. the event hasn't been updated).

Bug: T206576


app/Resources/assets/js/eventshow.js
```diff
@@ -141,3 +142,19 @@ eventmetrics.eventshow.setupWikiInputs = function () {
...
+eventmetrics.eventshow.setupReportModal = function () {
+    $('#download-modal').on('shown.bs.modal', function (event) {
+        var modal = $(this),
+            reportLink = $(event.relatedTarget);
+        modal.find('.modal-title').text(reportLink.data('modal-title'));
+        modal.find('.wikitext-link').attr('href', reportLink.data('wikitext-link'));
+        modal.find('.csv-link').attr('href', reportLink.data('csv-link'));
+        // Also append the event status, duplicated from the underlying page.
+        modal.find('.event-status').html($('.event-stats-status').html());
+    })
```
MusikAnimal: The one problem I see here is this happens after the modal is loaded, so the user sees the changes to the content as it happens. Just a little bit weird. Could we hook into `show.bs.modal` instead of `shown`?

i18n/en.json
```diff
@@ -64,9 +64,13 @@
 	"error-wikis": "$1 {{PLURAL:$1|wiki|wikis}} {{PLURAL:$1|is|are}} invalid. Check the spelling.",
 	"event": "Event",
 	"event-all-edits": "All edits",
+	"event-all-edits-list": "All edits list",
```
MusikAnimal: I talked to Joe and he said we can simply use the existing "All edits" message (`event-all-edits`). The ticket has been updated accordingly

samwilson: Oh cool. Fixed.

app/Resources/views/events/_header.html.twig
```diff
@@ -50,6 +73,26 @@
     </h1>
 </div>
 
+{### DOWNLOAD REPORT MODAL ###}
+<div id="download-modal" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog" role="document">
```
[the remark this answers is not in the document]
samwilson: Done.

app/Resources/assets/js/eventshow.js
samwilson: Oops, that's what I meant to do! Looks much better. Thanks.

i18n/en.json
```diff
@@ -64,9 +64,13 @@
...
+	"event-all-edits-report": "'All edits' report",
 	"event-created": "The event '$1' was successfully created.",
 	"event-deleted": "The event '$1' was deleted.",
+	"event-download-csv": "Download CSV",
 	"event-download-reports": "Download reports",
+	"event-download-wikitext": "Get Wikitext",
```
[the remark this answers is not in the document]
samwilson: Done.

MusikAnimal:
Good stuff! Looks like for https://phabricator.wikimedia.org/T206576 we still need to do the "Animated 'Working' Indicator—labeling and behavior" stuff. That may require further discussion... as the reports currently don't invoke an update and I think this requirement may really complicate things. We'll worry about that later :)

PR #216, opened 2019-03-08
Improve logging in EventProcessor, using Stopwatch to log runtimes

MusikAnimal:
This doesn't include actual query runtimes, as requested by T205322, but
it does reveal which pieces of the update process are the slowest.

Bug: T205322


MusikAnimal:
> I dunno if we need to round milliseconds to two decimal places. But all good. :)

Ha! You are correct. I could have sworn this was seconds. I'll change that later

PR #217, opened 2019-03-12
Add support for categories without participants

MusikAnimal:
Add support for categories without participants

This can still fail very large categories, but we'll iron out performance
issues later. Most categories process just fine without any error.

Some metrics don't apply when there are no participants, such as local
file uploads (there could be thousands, most presumably unrelated to the
event). To address this, we introduce EventWiki::canHaveFilesUploaded()
and also EventWiki::isValid(). The latter ensures there are at least
some participants or one category to work with.

Remove unused EventController::getEventStats().

Other minor code cleanup.

Bug: T205734

### DIFF
--- a/app/Resources/views/events/show.html.twig
+++ b/app/Resources/views/events/show.html.twig
@@ -42,7 +42,13 @@
                                             <span class="glyphicon glyphicon-info-sign" data-toggle="tooltip" data-placement="top" title="{{ msg(stat.metric ~ '-desc', [stat.offset]) }}"></span>
                                         {% endif %}
                                     </th>
-                                    <td>{{ stat.value|num_format }}</td>
+                                    <td>
+                                        {% if stat.value is null %}
+                                            &ndash;
+                                        {% else %}
+                                            {{ stat.value|num_format }}
+                                        {% endif %}
+                                    </td>
                                 </tr>
                             {% endif %}
                         {% endfor %}
@@ -61,9 +67,8 @@
                 <div class="text-center event-wiki-stats--empty">
                     {% if event.hasJob %}
                         &nbsp;
-                    {% elseif event.numParticipants == 0 %}
-                        {# FIXME: ^ also check for event.numCategories == 0 and add msg('categories') once T194707#4620358 is resolved #}
-                        {{ msg('event-misconfig', [msg('participants')]) }}
+                    {% elseif event.numParticipants == 0 and event.numCategories == 0 %}
+                        {{ msg('event-misconfig', [[msg('participants'), msg('categories')]|join(', ')]) }}
                     {% elseif event.startUTC > date() %}
                         {{ msg('event-in-future') }}
                         (<a href="{{ path('EditEvent', {'programId': program.id, 'eventId': event.id}) }}">{{ msg('settings')|lower }}</a>)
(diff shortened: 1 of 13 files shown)


src/AppBundle/Service/EventProcessor.php
```diff
@@ -535,18 +538,17 @@ private function setUserCounts(EventWiki $wiki, EventWikiRepository $ewRepo): vo
      */
     private function getParticipantNames(): array
     {
-        // Quick cache.
-        static $parUsernames = null;
-        if (null !== $parUsernames) {
-            return $parUsernames;
```
MusikAnimal: FYI @MaxSem , this was causing erroneous results for https://github.com/wikimedia/eventmetrics/pull/217/files#diff-421e5b20693d341d563a6dec2f088b95R441. Looking at that test, indeed the new editor and retention metrics should have been zero. The cheap "quick cache" code above was making it pull in the usernames from the previous test. My fault.

src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -444,8 +449,8 @@ public function getUsersFromPageIDs(string $dbName, array $pageIds, Event $event
             ->andWhere('rev_user <> 0')
             ->andWhere('rev_timestamp BETWEEN :start AND :end')
             ->setParameter('pageIds', $pageIds, Connection::PARAM_INT_ARRAY)
-            ->setParameter('start', $event->getStart()->format('YmdHis'))
-            ->setParameter('end', $event->getEnd()->format('YmdHis'));
+            ->setParameter('start', $event->getStartUTC()->format('YmdHis'))
+            ->setParameter('end', $event->getEndUTC()->format('YmdHis'));
```
MusikAnimal: Note this also. Should have been using UTC all this time.

src/AppBundle/Service/EventProcessor.php
MusikAnimal: To be clear, I wrote the bad caching code, you wrote the test which was asserting the wrong result. Definitely my fault, just thought I'd share this discovery.

MusikAnimal:
This should probably depend on #218, so that we can write better tests. Let's wait for that to get merged first, and I will update this PR accordingly.

app/Resources/views/events/show.html.twig
```diff
@@ -61,9 +67,8 @@
                 <div class="text-center event-wiki-stats--empty">
                     {% if event.hasJob %}
                         &nbsp;
-                    {% elseif event.numParticipants == 0 %}
-                        {# FIXME: ^ also check for event.numCategories == 0 and add msg('categories') once T194707#4620358 is resolved #}
-                        {{ msg('event-misconfig', [msg('participants')]) }}
+                    {% elseif event.numParticipants == 0 and event.numCategories == 0 %}
+                        {{ msg('event-misconfig', [[msg('participants'), msg('categories')]|join(', ')]) }}
```
samwilson: Not that it matters for this commit, but in https://github.com/wikimedia/ToolforgeBundle/pull/13 I'm adding a `list_format` Twig function that'll help with lists like this.

MusikAnimal: Good to know! I think this line of code will get removed with #234 anyway.

PR #218, opened 2019-03-13
Apply category filtering on Commons to files

MaxSem:
Bug: https://phabricator.wikimedia.org/T214744


i18n/qqq.json
```diff
@@ -22,7 +22,7 @@
 	"calculate-totals": "Label for button to calculate the statistics for an event.",
 	"cancel": "Label for button\n{{Identical|Cancel}}",
 	"categories": "Label for section showing MediaWiki categories associated with an Event.\n{{Identical|Category}}",
-	"category-description": "Description of how categories are used to generate statistics.",
+	"categories-description": "Description of how categories are used to generate statistics.",
```
MusikAnimal: qqq for `categories-commons-explanation`?

src/AppBundle/Repository/EventRepository.php
```diff
@@ -172,7 +185,7 @@ public function getPagesUsingFile(string $dbName, string $filename): int
             $rqb->select('COUNT(DISTINCT(CONCAT(gil_wiki, gil_page)))')
                 ->from('commonswiki_p.globalimagelinks')
                 ->where('gil_to = :filename')
-                ->andWhere('gil_page_namespace_id = 0');
+                ->andWhere('gil_page_namespace_id = 6');
```
MusikAnimal: This query looks at global file usage. We only want to count mainspace pages, right? I think this is what's causing the build to fail.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -236,14 +263,51 @@ private function getFileUsageBuilder(string $dbName, DateTime $start, DateTime $
...
+     * @param string[] $categories
+     */
+    private function applyUsersOrCategories(
+        QueryBuilder $rqb,
+        string $dbName,
+        array $usernames,
+        array $categories
+    ): void {
+        if ($usernames && $categories) {
+            // throw new InvalidParameterException('Cannot process events with both users and categories');
```
MusikAnimal: Code is commented out; but anyway we do allow participants + categories, so I don't think we need this check.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -236,14 +263,51 @@ private function getFileUsageBuilder(string $dbName, DateTime $start, DateTime $
...
+        $namespace = 'commonswiki_p' === $dbName ? 6 /* NS_FILE */ : 0 /* NS_MAIN */;
+        if ($usernames) {
+            $rqb->andWhere("img_user_text IN (:usernames)")
+                ->setParameter('usernames', $usernames, Connection::PARAM_STR_ARRAY);
+        } elseif ($categories) {
+            $rqb->join(
+                "$dbName.image",
+                "$dbName.page",
+                null,
+                "img_name = page_title AND page_namespace = /* NS_FILE */ $namespace"
```
MusikAnimal: Guessing you didn't mean for `/* NS_FILE */` to be here, since it's not necessarily the file namespace

src/AppBundle/Repository/EventRepository.php
MaxSem: D'oh!

src/AppBundle/Repository/EventRepository.php
MaxSem: Removed.

src/AppBundle/Repository/EventRepository.php
MaxSem: Done.

i18n/qqq.json
MaxSem: Done.

PR #221, opened 2019-03-14
Add dots animation to Update button

samwilson:
Add three pulsing dots to the 'queued' state of the Update button.

Bug: T206576


MusikAnimal:
Looks great! I can only assume the vertical placements of the dots in the [mock](https://phabricator.wikimedia.org/F27853296) was an oversight. You have it perfectly aligned in the middle, which looks fantastic.

PR #222, opened 2019-03-14
Store files uploaded as their own type of page IDs, update queries

MusikAnimal:
This simplifies the code and makes the queries faster. Additionally, the
revision table retains the original uploader of the file, while going
off of the image table we only get the most recent version (which may be
by a non-participant).

On the backend, the "type" of page ID is called "pagesFiles" -- the name
chosen so it can use the helpers for "pagesCreated" and "pagesImproved"
that reference class property by the given 'type'.

Only use revision_userindex when filtering by user.


tests/AppBundle/Repository/EventRepositoryTest.php
```diff
@@ -30,22 +30,4 @@ public function testGetPagesUsingFile():void
...
-        $ret = $this->repo->getPagesUsingFiles(
-            'commonswiki_p',
-            new \DateTime('2014-12-01'),
-            new \DateTime('2014-12-02'),
-            ['Krupin.1'],
-            []
-        );
-
-        static::assertCount(1, $ret);
-    }
```
MusikAnimal: This test isn't really needed anymore since `getPagesUsingFiles` lacks any complex logic. The integration tests will still cover the relevant part of the code.

PR #223, opened 2019-03-14
[WIP] Allow stat values to be null, display as a ndash in the UI.

MusikAnimal:
Some metrics we report 0 when it's really a N/A. With this commit
these metrics can be saved as null. This allows us to replace the
confusing 0's in the UI to a simple &ndash;


MusikAnimal:
Abandoning. I don't think we need this.

PR #224, opened 2019-03-14
Kill active database connections between tests

MusikAnimal:
Apparently they persist, which can cause you to hit the 5-connection
limit imposed on the replicas. This issue is intermittent.

Also throw Exception in JobHandler when db connection limit is hit,
so that it's more clear what went wrong.


MusikAnimal:
PHPUnit and/or CodeClimate is ignoring the `@codeCoverageIgnoreStart` annotation. You can disregard that failure

PR #225, opened 2019-03-15
Allow Wiktionary and Wikivoyage events

MaxSem:
Bug: https://phabricator.wikimedia.org/T217058


samwilson:
This makes the new wiki families appear in the list, but doesn't let you save them: I get "1 wiki is invalid. Check the spelling."

MusikAnimal:
Eek, so there's also:
* https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/Service/EventProcessor.php#L266-L267
* https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/Model/Event.php#L68 (just copy/paste to create new entries for Wiktionary/Wikivoyage for now, later we could do some refactoring)

I _think_ that's it. It may be worthwhile to add a test case for at least one non-Wikipedia. You could add to the existing fixtures, or create a new one. I would test for something very simple like the total number of edits (`edits` metric). We just want to make sure a non-Wikipedia successfully makes it way through the pipeline.

MaxSem:
Still needs tests.

MusikAnimal:
Such beautiful code! Works well in my testing.

PR #228, opened 2019-03-15
Reorganise the metadata part at the top of the event page

samwilson:
Move existing items (dates and wiki list) around, and add new
category list. Both lists are now formatted with a new
'list_format' Twig filter which uses Intuition's built-in
list method.

Bug: T218340


MaxSem:
The commit looks good, however I have a question about the spec, asked it in the bug itself.

MusikAnimal:
Surely this isn't the intended appearance? It also looks odd to have "Category:" when there are none.

Screenshot from 2019-03-20 17-28-02

samwilson:
I think it should have a dash. I've asked on the ticket for clarification. This whole patch wasn't actually required it turns out. Oops.

PR #234, opened 2019-03-19
Reconfigure filter warning messages

samwilson:
Set up three filter-message sections (only one of which is
displayed) and modify the form-panel descriptions depending
on the event state.

Bug: T218340


src/AppBundle/Controller/EventController.php
```diff
@@ -199,6 +199,7 @@ public function showAction(JobHandler $jobHandler): Response
             $forms[$formType] = $form->createView();
         }
 
+        $filtersMissing = 0 === $this->event->getNumParticipants() && 0 === $this->event->getNumCategories(true);
```
MaxSem: There are 7 equals signs on this line, not readable if you ask me.

samwilson: :) yes it does look a bit odd doesn't it. I've changed it to use implicit casts.

MusikAnimal: I would probably put this in the `Event` class, and just reference it in the view as with `event.filtersMissing`. This is because it sort of ties into `Event::isValid()`.

On that note I wonder if we should hold off on this until #217 is merged, which changes some of this logic.

src/AppBundle/Model/Event.php
```diff
@@ -617,6 +620,31 @@ public function getWikis(): Collection
...
+            }
+            foreach ($this->categories as $cat) {
+                if ($cat->getDomain() === $eventWiki->getDomain()) {
+                    return false;
+                }
+            }
+            return true;
+        });
+        return $this->wikisWithoutCategories;
+    }
```
MusikAnimal: I think this method deserves a unit test

app/Resources/views/events/show.html.twig
```diff
@@ -26,6 +26,71 @@
...
+        </section>
+
+        {% else %}
+            <div class="alert alert-success">
+                {% if event.startUTC > date() %}
+                    {{ msg('event-in-future') }}
+                    (<a href="{{ path('EditEvent', {'programId': program.id, 'eventId': event.id}) }}">{{ msg('settings')|lower }}</a>)
+                {% else %}
+                    {{ msg('event-no-data') }}
+                    <a href="#" class="event-process-link">{{ msg('calculate-totals') }}</a>.
```
MusikAnimal: After I click "Calculate totals" the message does not disappear (which it did before).

samwilson: Fixed. Now checks for job (as it did before).

MusikAnimal:
Another possible bug (sorry for the flurry of comments): If I have two wikis, en.wikipedia and commons.wikimedia, I still see "Event partially configured" even though it should be valid. Perhaps this is because #217 hasn't been merged yet? Overall I think this PR should wait on it... what do you think?

app/Resources/views/events/show.html.twig
```diff
@@ -26,6 +26,71 @@
...
+                    </strong>
+                </li>
+                <li>{{ msg('error-filters-every-wiki') }}</li>
+                {% if event.wikiByDomain('www.wikidata') %}
+                <li>{{ msg('error-filters-wikidata') }}</li>
+                {% endif %}
+                <li>{{ msg('error-filters-participants-desc') }}</li>
+                <li>{{ msg('error-filters-categories-desc') }}</li>
+            </ul>
+        </section>
```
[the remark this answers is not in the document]
samwilson: Done.

I guess there's no linter for the Twig standards?

app/Resources/views/events/show.html.twig
```diff
@@ -26,6 +26,71 @@
...
         {% include 'events/_metadata.html.twig' with {'event': event} %}
 
+        {% if filtersMissing %}
+        <section class="event-errorbox error">
+            <h3>{{ msg('error-filters-header') }}</h3>
+            <p>
+                {{ msg('error-filters-desc') }}
+                <a href="https://meta.wikimedia.org/wiki/Special:MyLanguage/Event_Metrics#adding-participants"
+                   title="{{ msg('error-filters-desc-tooltip') }}">
+                    <span class="glyphicon glyphicon glyphicon-question-sign"></span>
```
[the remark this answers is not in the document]
samwilson: Fixed.

app/Resources/views/events/show.html.twig
```diff
@@ -26,6 +26,71 @@
...
+                    {{ msg('error-filters-categories-missing') }}
+                    <strong>
+                        {% for noCatWiki in wikisWithoutCats %}
+                            {{ noCatWiki.domain }}{% if not loop.last %},{% endif %}
+                        {% endfor %}
+                    </strong>
+                </li>
+                <li>{{ msg('error-filters-every-wiki') }}</li>
+                {% if event.wikiByDomain('www.wikidata') %}
+                <li>{{ msg('error-filters-wikidata') }}</li>
```
MusikAnimal: Indentation

i18n/en.json
```diff
@@ -48,6 +52,18 @@
 	"error-event-retry-link": "report this issue",
 	"error-event-timeout": "This process took too long and was stopped.",
 	"error-event-undeletable": "This event cannot be deleted because it has one or more participants.",
+	"error-filters-categories-desc": "Category filters are specific to individual wikis.",
+	"error-filters-categories-missing": "The following wikis require filtering:",
+	"error-filters-combination": "If you combine Categories and Participants, the system will provide metrics for contributions that satisfy both filters (i.e., Category AND Participants).",
+	"error-filters-desc": "Event setup isn’t complete yet. Please use one or both of the filters below to more narrowly focus your metrics.",
```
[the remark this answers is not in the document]
samwilson: Oh yeah... I left it in because it's better typography, but consistency is even better that that. I've made em straight.

i18n/en.json
```diff
@@ -48,6 +52,18 @@
...
+	"error-filters-desc-tooltip": "Read more about how to enter participant usernames",
+	"error-filters-every-wiki": "You must provide filtering for every wiki for which you want metrics.",
+	"error-filters-header": "Filters required",
+	"error-filters-participants-desc": "Participants filtering applies to all wikis and satisfies all filtering requirements.",
+	"error-filters-participants-none": "No participants supplied",
+	"error-filters-participants-or-categories": "You can filter by Participants or Categories—or combine the two.",
+	"error-filters-wikidata": "To get metrics about <strong>Wikidata</strong>, you must supply Participants. (Categories don’t exist on Wikidata.)",
```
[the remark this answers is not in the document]
samwilson: Yeah, the spec says "Please observe boldfacing", and we don't like piecemeal messages. We *could* lookup the name Wikidata in the language, and inject it, but even then it's probably not the right thing to do from a l10n perspective.

I'm always confused about the best way to do these things. I wish we could have messages in Markdown!

i18n/en.json
```diff
@@ -129,6 +148,7 @@
 	"pageviews-average": "Avg. daily pageviews",
 	"pageviews-cumulative": "Pageviews, cumulative",
 	"participants": "Participants",
+	"participants-panel-title": "Participants filter",
 	"please-login": "Please login to continue.",
 	"program": "Program",
 	"program-created": "The program '$1' was successfully created.",
```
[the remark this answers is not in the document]
samwilson: Removed.

i18n/qqq.json
```diff
@@ -22,8 +22,12 @@
 	"calculate-totals": "Label for button to calculate the statistics for an event.",
 	"cancel": "Label for button\n{{Identical|Cancel}}",
 	"categories": "Label for section showing MediaWiki categories associated with an Event.\n{{Identical|Category}}",
+	"categories-panel-title": "Title for the categories' filter form panel. Same function as {{msg|participants-panel-title}}.",
```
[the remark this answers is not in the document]
samwilson: Oops! I've made this mistake before. :( Fixed (replaced with `{{msg-wm}}`).

src/AppBundle/Model/Event.php
```diff
@@ -208,6 +208,9 @@ class Event
      */
     protected $jobs;
 
+    /** @var ArrayCollection */
+    protected $wikisWithoutCategories;
```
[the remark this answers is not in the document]
samwilson: Good point. Removed.

src/AppBundle/Model/Event.php
```diff
@@ -617,6 +620,31 @@ public function getWikis(): Collection
         return $this->wikis;
     }
 
+    /**
+     * Get all of this event's wikis that do not yet have at least one category defined.
+     * Wikidata is excluded because it can never have categories.
+     * @return ArrayCollection Collection of EventWiki objects.
+     */
+    public function getWikisWithoutCategories(): ArrayCollection
```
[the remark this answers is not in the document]
samwilson: Done.

src/AppBundle/Model/Event.php
samwilson: Done.

samwilson:
> If the event has had at least one update, I still see "Statistics have not yet been generated for this event. Calculate totals." in a green `alert-success` banner at the top. That seems like a bug, or maybe something's wrong on my end? Either way I think should be `alert-warning`, and also maybe add a bit of spacing above the banner:
> 
> Also the new banners are missing some spacing between it and the metadata above it. I don't think it's need to be as much as space as in the mock, but it looks a little odd now:
> 

I was working on the idea that #228 would be done as well, but now that it's not I'll update these. The green box however is in the spec (although the message needs to be updated; I was going to do that in a separate patch along with some other message changes for the two forms — but part of the same task).

I've added some spacing below the metadata.

app/Resources/views/events/show.html.twig
samwilson: Done.

samwilson:
> Another possible bug (sorry for the flurry of comments): If I have two wikis, en.wikipedia and commons.wikimedia, both with categories, I still see "Event partially configured" even though it should be valid. Perhaps this is because #217 hasn't been merged yet? Overall I think this PR should wait on it... what do you think?

Oops, I was failing to actually *count* the wikisWithoutCats... done now.

samwilson:
Okay, I think this is ready for re-review.

src/AppBundle/Model/Event.php
```diff
@@ -616,6 +616,27 @@ public function getWikis(): Collection
         return $this->wikis;
     }
 
+    /**
+     * Get all of this event's wikis that do not yet have at least one category defined.
+     * Wikidata is excluded because it can never have categories.
+     * @return Collection Collection of EventWiki objects.
+     */
+    public function getWikisWithoutCategories(): ArrayCollection
```
MusikAnimal: The return type declaration also needs to be `Collection`. I think if there are EventWikis that are persisted but haven't been flushed to the database, the `$this->wikis->filter` code will return a `PersistentCollection`, which in turn will cause an exception.

samwilson: Oops! Of course. :) Done.

samwilson:
> I'm still seeing the "Statistics have not yet been generated for this even" banner even if they have been generated. And it doesn't hide itself once the update starts. 

Fixed. It now hides itself when either of the 'update data' links are clicked.

> I also think whatever banner is at the top really needs some padding above it. A bit too tight up against the metadata part. But we can take care of that later, maybe this was an intentional design choice.

I've given it 1em.

MusikAnimal:
Looks good to me! Just need to resolve the merge conflict

PR #237, opened 2019-03-20
Upgrade the ToolforgeBundle and make required changes

samwilson:
The new version of the bundle includes a bunch of new Twig
filters and functions. This change removes these from
Event Metrics and changes a bit of usage. The main new
functionality is to add in comma-separate list formatting that
is correct for all languages.


tests/AppBundle/Twig/AppExtensionTest.php
```diff
@@ -1,48 +0,0 @@
...
-
-    /**
-     * Methods that fetch data about the git repository.
-     */
-    public function testGitMethods(): void
-    {
-        static::assertEquals(7, strlen($this->appExtension->gitShortHash()));
-        static::assertEquals(40, strlen($this->appExtension->gitHash()));
-    }
-}
```
MusikAnimal: Did we want to move these tests to the bundle? I guess they aren't that important. I think I only wrote them to improve our test coverage stats :) Just a thought.

tests/AppBundle/Twig/FormatExtensionTest.php
```diff
@@ -97,16 +105,6 @@ public function testFormatDuration(): void
...
-    /**
-     * Format a number.
-     */
-    public function testNumberFormat(): void
-    {
-        static::assertEquals('1,234', $this->formatExtension->numberFormat(1234));
-        static::assertEquals('1,234.32', $this->formatExtension->numberFormat(1234.316, 2));
-        static::assertEquals('50', $this->formatExtension->numberFormat(50.0000, 4));
-    }
-
```
MusikAnimal: Same here. These seem more meaningful and maybe worthy of moving to the bundle.

samwilson: Oh oops, yes I totally remembered that earlier today but forgot to push the code. It's at https://github.com/wikimedia/ToolforgeBundle/pull/14 now.

PR #241, opened 2019-03-22
Exclude pages created from pages improved and rename the entity property

MusikAnimal:
This was semi-fixed before, but still broke if a participant edited the
page after creating it.

Writing a query that looked for unique pages edited, but not created, by
a list of users and/or within a category, turned out too be a mess and
possibly infeasible. Instead, we fetch the pages created then edited,
and subtract the arrays. This logic is handled in EventProcessor.

Other minor code refactoring.

Bug: T217455


MusikAnimal:
Pweez ignore CodeClimate. I don't see what is missing tests here

PR #246, opened 2019-03-26
Add more filtering-display logic

samwilson:
Add the remaining filter-display logic for the event page.

This also adds a new parameter for the form panel macro, to
toggle its initial collapsed state.

Bug: T218340


app/Resources/views/events/show.html.twig
```diff
@@ -26,44 +26,54 @@
...
-                                {{ noCatWiki.domain }}{% if not loop.last %},{% endif %}
-                            {% endfor %}
-                        </strong>
+                        {% set cats = [] %}
+                        {% for noCatWiki in wikisWithoutCats %}
+                            {% set cats = cats|merge(['<strong>'~noCatWiki.domain~'</strong>']) %}
+                        {% endfor %}
+                        {{ cats|list_format|raw }}
                     </li>
+                    {% endif %}
```
MusikAnimal: Not important but I would indent the content within the if/endif

PR #249, opened 2019-03-27
Change participants and categories form labels

samwilson:
Add extra lines with more user instructions, and link to on-wiki
documentation (via fragment IDs that are already added as
{{anchor}}s on that page).

Bug: T218340


app/Resources/views/events/_categories.html.twig
```diff
@@ -29,8 +29,17 @@
     </div>
 
     {{ form_start(form, {'method': 'post', 'name': 'categoryForm', 'attr': {'class': 'save-categories-form', 'autocomplete': 'off'}}) }}
-    <p class="save-categories-form__description">{{ msg('categories-description') }}</p>
-    <p>{{ msg('categories-commons-explanation') }}</p>
+    <p class="save-categories-form__description">
+        {{ msg('categories-description') }}
+        <a href="https://meta.wikimedia.org/wiki/Special:MyLanguage/Event_Metrics#category-filtering"
+           title="{{ msg('categories-description-tooltip') }}" class="help-link" target="_blank">
+            <span class="glyphicon glyphicon glyphicon-question-sign"></span>
```
MusikAnimal: two `glyphicon`s

i18n/en.json
```diff
@@ -24,6 +25,8 @@
 	"categories": "Categories",
 	"categories-panel-title": "Categories filter",
 	"categories-description": "Enter categories (with wiki names) to limit metrics to pages in those categories",
+	"categories-description-tooltip": "Learn about Categories filtering",
```
MusikAnimal: looks like we're missing qqq for this message

app/Resources/views/events/_categories.html.twig
samwilson: I've made this mistake before! :)

Fixed.

i18n/en.json
samwilson: Fixed.

MusikAnimal:
I ran the tests locally and they passed.

PR #250, opened 2019-03-27
Add pages improved wikitext report

MaxSem:
Also fix an always plus issue in pages created.

Bug: https://phabricator.wikimedia.org/T210898


MusikAnimal:
@MaxSem On an unrelated and unimportant note, we've putting `Bug: T123` in commit summary, since this will make the commit show up in the ticket on Phabricator. A little annoying because it's not a link here on GitHub, but whatevs.

PR #251, opened 2019-03-28
Count Wikidata edits, update messaging and logging

MusikAnimal:
Improve test coverage

Bug: T218445


MusikAnimal:
I don't Joe updated T218445 yet but I talked to him offline about including Wikidata and we decided to do it. This was supposed to be a one-line change (aside from the update to the messages, see [T218445#5063587](https://phabricator.wikimedia.org/T218445#5063587)), but I saw an opportunity to do a little refactoring so I did. I also improved the logging to be more clear about where we are in the process.

samwilson:
(Oops, pressed 'comment' too quickly.)

I meant to say: there's duplication between 'pages' and 'items', and the former shouldn't be there at all if it's just for Wikidata.

MusikAnimal:
> there's duplication between 'pages' and 'items', and the former shouldn't be there at all if it's just for Wikidata.

Fixed! I went with just "pages created/improved", hopefully it's clear to devs that in this case we're talking about items.

PR #252, opened 2019-03-28
Don't collapse categories/participants on error

MaxSem:
Bug: https://phabricator.wikimedia.org/T219464


MusikAnimal:
Doesn't show errors when the wiki for a category is invalid, but that has nothing to do with this PR. Merging!

PR #257, opened 2019-03-29
Change future-event warning to orange box style

samwilson:
Bug: T219436


MaxSem:
Rebased, tests pass locally.

PR #258, opened 2019-03-29
Add green checkmarks to form panels

samwilson:
Also include Bootstrap's standard 'success' green for the
alert boxes.

Bug: T218340


MaxSem:
I see the checkmark even when not everything's alrite:
Screen Shot 2019-03-29 at 14 00 19

samwilson:
> I see the checkmark even when not everything's alrite:

Good point. I've fixed it to not show the check mark or participants/categories count if there are invalid form fields. It could probably also show an orange icon there, but there's already plenty of alerts for that stuff, and this is easier.

PR #259, opened 2019-03-29
Don't display "Event settings saved successfully" on errors

MaxSem:
Bug: T219464


samwilson:
I was going to say that the `hasFormErrors` could probably be more cleanly calculated in the controller, where we're already looping through the forms. But that's a minor detail. :)

PR #261, opened 2019-03-29
Fix fetching of average pageviews for Pages Improved report

MusikAnimal:
Avg pageviews are always a 31-day window, and not since the start
of the event.

Bug: T219650

### DIFF
--- a/src/AppBundle/Repository/EventWikiRepository.php
+++ b/src/AppBundle/Repository/EventWikiRepository.php
@@ -9,7 +9,6 @@ namespace AppBundle\Repository;
 
 use AppBundle\Model\Event;
 use AppBundle\Model\EventWiki;
-use DateInterval;
 use DateTime;
 use Doctrine\DBAL\Connection;
 use Exception;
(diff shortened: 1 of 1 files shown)


src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -293,13 +292,9 @@ public function getPageviews(
 
         $pageviewsRepo = new PageviewsRepository();
         $recentDayCount = Event::AVAILABLE_METRICS['pages-improved-pageviews-avg'];
-
-        // The offset date is the start of the period over which pageviews should be averaged per day, up to today.
-        $offsetDate = (new DateTime('yesterday midnight'))->sub(new DateInterval('P'.$recentDayCount.'D'));
-        $start = $getDailyAverage && $start < $offsetDate ? $offsetDate : $start;
```
MusikAnimal: This logic was moved to `PageviewsRepository::getAvgPageviewsPerArticle()` a while back and I forgot to remove it here. When `$getDailyAverage` is set then `$start` isn't used, so basically this code didn't do anything.

PR #262, opened 2019-03-29
Display an alert before navigating if the form is 'dirty'     

mooeypoo:
Make sure that if any form is in a dirty state, or if there
are any errors on a form, we ask the user whether they're  
certain they want to navigate away.

Make sure that this does not prevent submission of the form
and that we do not interrupt regular navigation.

Bug: https://phabricator.wikimedia.org/T212804


mooeypoo:
Okay, y'all are making a good point, I read the ticket and slightly misunderstood it, I thought the ask is stemming from the saved/unsaved status of the invalid usernames (that is then hard to notice because it comes back after you save, but it wasn't *actually* saved.

I'll go over this and try to spot both cases of invalid usernames as well as any "dirty" input state.

mooeypoo:
> Thanks for this! I think we should listen for changes to any input and set a dirty state JS variable, and check that in `window.onbeforeload`. If we put this in application.js, it'll apply to all forms across the application. You'd still need to check for invalid entities; for that I think you could just look for `.has-error` in the DOM.

Is .has-error including the invalid-username case? I wasn't 100% sure which class to find, but it sounds reasonable since that state does show the error box.

Good idea on generalizing it into `application.js`! I'll work on that.

app/Resources/assets/js/application.js
```diff
@@ -15,6 +15,55 @@ $(function () {
 
     // Activate Bootstrap tooltips.
     $('[data-toggle="tooltip"]').tooltip();
+
+    // For pages with forms, check if there are any "dirty"
+    // inputs and use that to notify the user before navigating away
+    if ( $('form').length > 0 ) {
```
MusikAnimal: phpcs isn't catching this, but we're not using the extraneous spacing within `if`s or arguments of a function, like here and on lines 39, 40, and 60. I will look into fixing the linter rules

MusikAnimal:
The "database quota exceeded" errors in the tests was probably because the connections weren't closed in the last build that was cancelled before it finished. You didn't change any PHP so I'm quite sure tests will pass, merging!

PR #269, opened 2019-04-01
Give full list of all unfiltered wikis

samwilson:
The list of unfiltered wikis was excluding Wikidata because it
does not have categories. This adds a new method to retrieve the
uncategorized wikis and Wikidat in addition if there are no
participants.

Bug: T218340

### DIFF
--- a/app/Resources/views/events/show.html.twig
+++ b/app/Resources/views/events/show.html.twig
@@ -68,14 +68,14 @@
                     {{ help_link }}
                 </p>
                 <ul>
-                    {% if wikisWithoutCats.count %}
+                    {% if wikis_without_filters.count %}
                     <li>
                         {{ msg('error-filters-categories-missing') }}
-                        {% set cats = [] %}
-                        {% for noCatWiki in wikisWithoutCats %}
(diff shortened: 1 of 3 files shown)


MusikAnimal:
When I have say en.wikivoyage and www.wikidata configured, with no categories or participants, I used to get a long list of messages:
```
To get metrics about Wikidata, you must supply Participants. (Categories don't exist on Wikidata.)
You can filter by Participants or Categories—or combine the two.
You must provide filtering for every wiki for which you want metrics.
Participants filtering applies to all wikis and satisfies all filtering requirements.
Category filters are specific to individual wikis.
If you combine Categories and Participants, the system will provide metrics for contributions that satisfy both filters (i.e., Category AND Participants).
```

With this PR I see only:
```
To get metrics about Wikidata, you must supply Participants. (Categories don't exist on Wikidata.)
```

Is this intentional?

In neither case do I see the "The following wikis require filtering: ..." message.

samwilson:
Yeah, that's right. The list of unfiltered wikis is only shown on the *Partial* state (orange message box), because in the *Unfiltered* state (red box) there's no need to tell the user which bits are unfiltered because everything is.

#272 is fixing which messages are displayed.

PR #271, opened 2019-04-01
Replace the language dropdown with ULS

mooeypoo:
Load ULS with a dedicated language button, and reload the
page with ?uslang=chosenLanguage. Also move the 'translate
this' help link into the help menu (this part was validated
with product).

Bug: https://phabricator.wikimedia.org/T219746


mooeypoo:
I don't think we can come up with a really good way to sort languages, especially ones that are non-latin properly. There are some ways out there, like what MediaWiki uses for the sidebar, but they're not great. ULS comes with the power of enabling the user to quickly find their language, *and* give us details that we need for RTL support (Which, btw, we should set up...)

I think the ideal would've been to get that language button to some settings page, and then we don't have to load it on all pageloads, but that would require a bit more work. We could reassess the ordering, but ULS (since we talked about it in the past) was a low-hanging fruit?

 I don't think the size is too horrible, but it *is* significant, especially for every page load. I'm not sure here.

MusikAnimal:
We've been down this road before ([T203699](https://phabricator.wikimedia.org/T203699)). I think we should focus on the task at hand and simply fix the sorting, which should be a [few-line change](https://github.com/x-tools/xtools/blob/master/src/AppBundle/Helper/I18nHelper.php#L140-L143), and save the ULS debate for Phabricator.

I admit I'm a bit stubborn about all this. I just do not see a world where some 100+ KB of JavaScript (twice the size of our own JS -- including jQuery, Moment, and Bootstrap, which we actually pull from a CDN) plus some 180+ files are needed for a language picker, at least in an application where this action wouldn't be commonplace. We should not feel compelled to use ULS just because our wikis do.

Intuition has a `isRTL()` function you could use to conditionally add a CSS override stylesheet, so I think we're covered there.

samwilson:
Do we have any idea how often people use the language selector? Or, rather, how often people use it more than once? Because it seems to me that a) it should quite often be defaulting to user's preferred language anyway because of the Accept-Language header; and b) if they do have to change it, we should remember that change (in a cookie) and they'll probably not need to change it very often.

Of course some people will need to change it regularly, for whatever reason, but it does feel less than efficient to have such a large part of every response be a big list of languages — even the status quo is too much I think.

That said, we don't want to go rebuilding too much of Event Metrics at this point (to e.g. add a Preferences page), but we could switch to ULS and only load it when they click the button (it's easy to add it as a separate asset target, so it'd just be four extra requests I think: the CSS, JS, a font, and an image). That doesn't work for non-JS but maybe that's okay.

Oh yeah, and the other thing that I wondered about: we could just retrieve the user's `language` preference from Meta and use that, and if they want to change it they can change it there.

MusikAnimal:
> Do we have any idea how often people use the language selector? Or, rather, how often people use it more than once?

I don't think so. Not for Event Metrics, anyway.

> a) it should quite often be defaulting to user's preferred language anyway because of the Accept-Language header; and b) if they do have to change it, we should remember that change (in a cookie) and they'll probably not need to change it very often.

This is my argument, that we don't need bloated software for something people will rarely use. Currently changes to your language are retained using a cookie.

As an aside: It's a bit a shame that that the cookie is tied to eventmetrics.wmflabs.org instead of just wmflabs.org, because if it were the latter your language preference would carry over to all other Intuition apps (Pageviews Analysis, XTools, SVG Translate (?), and many more). Not something we should be concerned about, just a thought.

> Of course some people will need to change it regularly, for whatever reason, but it does feel less than efficient to have such a large part of every response be a big list of languages — even the status quo is too much I think.

The HTML-only list I don't think is that huge, but indeed it isn't really needed probably 99% of the time.

> That said, we don't want to go rebuilding too much of Event Metrics at this point (to e.g. add a Preferences page), but we could switch to ULS and only load it when they click the button (it's easy to add it as a separate asset target, so it'd just be four extra requests I think: the CSS, JS, a font, and an image). That doesn't work for non-JS but maybe that's okay.

Good idea! My biggest argument against ULS is the size, so really like this approach. We'd need a loading indicator of sorts, methinks.

I still have some other less compelling arguments against ULS:
* It shows every language conceivable (seemingly). This could be misleading because most of those won't have any translations.
* The "Help translate" link should be somewhere in the ULS popup, preferably. I wouldn't expect it to be under "Help" (which is about documentation). I think this is doable, we'd just add some tiny JS to insert a link somewhere in the dialog.
* Without significant effort, you are stuck with ULS's styling. Probably not a big deal for us; again it's not like people change their language that often, but it'd be neat to have some control and keep the styling consistent with the rest of the application.

> Oh yeah, and the other thing that I wondered about: we could just retrieve the user's `language` preference from Meta and use that, and if they want to change it they can change it there.

Interesting idea. A lot of folks probably don't go to Meta that much or at all, but I assume it like Event Metrics goes off of the Accept-Language header. I doubt they are often different; if I prefer Spanish I would probably want Spanish everywhere on the web and will have this set as the default in my browser. I don't think they should have to go to Meta to change their language, though.

----

Overall, if we only load ULS when needed, I'm okay with it :) But I'll remind you addressing the task at hand ([T219746](https://phabricator.wikimedia.org/T219746)) at its simplest should be a [four-line change](https://github.com/x-tools/xtools/blob/master/src/AppBundle/Helper/I18nHelper.php#L140-L143). So you may wish to weigh out cost/benefit (though it seems like the bulk of the work has been done with this PR).

mooeypoo:
As per discussion (see ticket) the strategy is first to reorder the language list and not add ULS.
We should, at some point, consider doing a language audit on EventMetrics where we consider adding proper RTL support as well as language dropdowns in a settings page, in which case ULS may be reconsidered to be added.

This is not a priority, though. Closing this.

PR #279, opened 2019-04-05
PageviewsRepository: introduce HTTP retries and timeouts

MaxSem:
Prevent temporary network glitches from slowing down the process unduly.
At the same time, retry in case the problem was a one-off.

Bug: T219897


MusikAnimal:
With this PR, are you now seeing retries/failures being logged, or is this just a theoretical scenario?

MaxSem:
Nah, it's just an untested sketch that I had to fit into the spike.

PR #280, opened 2019-04-08
Mark long running jobs as timed out

MaxSem:
...instead of nuking them outright.

Bug: T220423

### DIFF
--- a/src/AppBundle/Service/JobHandler.php
+++ b/src/AppBundle/Service/JobHandler.php
@@ -9,6 +9,7 @@ namespace AppBundle\Service;
 
 use AppBundle\Model\Event;
 use AppBundle\Model\Job;
+use DateTime;
 use Doctrine\DBAL\Connection;
 use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/JobHandler.php
```diff
@@ -131,9 +133,15 @@ public function handleStaleJobs(Event $event): void
...
 
+        $dayAgo = new DateTime('-1 day');
         /** @var Job $job */
         foreach ($staleJobs->getIterator() as $job) {
-            $event->removeJob($job);
+            if ($job->isBusy()) {
+                $job->setStatus(Job::STATUS_FAILED_TIMEOUT);
+            }
+            if ($job->getSubmitted() >= $dayAgo) {
+                $event->removeJob($job);
```
MusikAnimal: `Job::isBusy()` gets Jobs that are either queued or started. The only other states are failed states, those are left intentionally so the user can see them. So I guess we don't need to remove the Job at all.

MaxSem: This just makes sure that very old jobs don't get stored forever.

MaxSem:
I don't think extra efforts would help here. The purpose of this ticket is to make jobs that likely have problems to become visibly timed out. From a product perspective, the decision is to consider jobs timed out after one hour, so I think that extra semantic correctness wouldn't help here.

src/AppBundle/Service/JobHandler.php
MusikAnimal: Stored forever is fine, I think... If I leave a job running, and come back the next day, I might wonder why nothing happened even though I recall having initiated an update. I don't have any strong feelings, just a thought.

MusikAnimal:
> There is one other concern to comes to mind. While a job is in the queued or started states, what happens when I try to manually browse to the URLs for the revision browser and reports? I think they should redirect back to the Event Summary page, since the data produced by those reports needs the Event update to finish (since it gives the page IDs). This can be done in a separate PR, though.

I created a PR for this at #283

PR #281, opened 2019-04-09
Bump time limit for jobs to 30 minutes

MaxSem:
Bug: T220463

### DIFF
--- a/src/AppBundle/Controller/EventDataController.php
+++ b/src/AppBundle/Controller/EventDataController.php
@@ -197,6 +197,9 @@ class EventDataController extends EntityController
             // JavaScript will poll for job status and update the view accordingly.
             $this->session->save();
 
+            // To process the job in this web request, bump execution time limit to 60 minutes.
+            set_time_limit(Job::TIME_LIMIT);
+
             // Attempt to start the job immediately (if there's quota).
             $jobHandler->spawn($job);
         }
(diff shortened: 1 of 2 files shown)


src/AppBundle/Controller/EventDataController.php
```diff
@@ -197,6 +197,9 @@ public function generateStatsAction(JobHandler $jobHandler): Response
             // JavaScript will poll for job status and update the view accordingly.
             $this->session->save();
 
+            // To process the job in this web request, bump execution time limit to 30 minutes.
+            set_time_limit(30 * 60);
```
MusikAnimal: Wouldn't we want to do an hour, to match `Event::getStaleJobs()`? I think the time constraints should live as constants in `Job`. We don't actually pass in anything to getStaleJobs() where it is used, so you could get rid of that argument and have that method use Job::MAX_LIFE (or whatever you want to call it). You'd use that constant here, too.

MusikAnimal:
Another thing... I will admit I haven't noticed this problem on my local. Jobs take more than 30 seconds and still finish. In fact, checking php.ini in production, I see `max_execution_time => 0 => 0`. Are we sure the PHP timeout is the culprit?

Feel free to `git fetch origin && sh deploy limits` on staging, and see if this patch makes a difference.

MusikAnimal:
> We don't actually pass in anything to getStaleJobs() where it is used, so you could get rid of that argument and have that method use Job::MAX_LIFE (or whatever you want to call it). You'd use that constant here, too.

Will take care of that in a separate PR.

I couldn't confirm this actually works, but I suppose it can't hurt! Merging.

MusikAnimal:
Oops, commit summary still says 30 minutes, doh well.

PR #284, opened 2019-04-10
Reconnect timed out database connections

MaxSem:
Tested by inserting a 15 minutes sleep.

Bug: T219897

### DIFF
--- a/src/AppBundle/Service/EventProcessor.php
+++ b/src/AppBundle/Service/EventProcessor.php
@@ -14,6 +14,8 @@ use AppBundle\Model\EventWikiStat;
 use AppBundle\Repository\EventRepository;
 use AppBundle\Repository\EventWikiRepository;
 use DateTime;
+use Doctrine\Bundle\DoctrineBundle\Registry;
+use Doctrine\DBAL\Connection;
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/EventProcessor.php
```diff
@@ -811,4 +817,20 @@ private function logEnd(string $key, ?string $message = null): void
...
+    private function reconnectDatabases(): void
+    {
+        /** @var Registry $registry */
+        $registry = $this->container->get('doctrine');
+        /** @var Connection $connection */
+        foreach ($registry->getConnections() as $connection) {
+            if (!$connection->ping()) {
+                $connection->close();
+                $connection->connect();
+            }
```
MusikAnimal: Why not move these checks to each connection method in `Repository`? https://github.com/wikimedia/eventmetrics/blob/master/src/AppBundle/Repository/Repository.php#L197-L259

There are other scenarios where this could happen, such as the reports. Anything that has a sequence of long-running queries.

MusikAnimal: I will say this is the exact approach I took before, and it didn't work for whatever reason. Hopefully it does this time!

PR #286, opened 2019-04-11
Fix incorrect message about files uploaded

samwilson:
Bug: T219379

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -163,7 +163,7 @@
 	"pages-improved-pageviews-avg": "Avg. daily views to pages improved",
 	"pages-improved-pageviews-avg-desc": "A 31-day average of daily pageviews to all Pages Improved during the event (compiled once per day). If fewer than 31 days are available, the average is calculated for the available days.",
 	"pages-using-files": "Unique pages with uploaded files",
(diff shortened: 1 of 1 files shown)


i18n/en.json
```diff
@@ -163,7 +163,7 @@
 	"pages-improved-pageviews-avg": "Avg. daily views to pages improved",
 	"pages-improved-pageviews-avg-desc": "A 31-day average of daily pageviews to all Pages Improved during the event (compiled once per day). If fewer than 31 days are available, the average is calculated for the available days.",
 	"pages-using-files": "Unique pages with uploaded files",
-	"pages-using-files-desc": "A count of all non-duplicate main namespace pages, on all wikis, on which Files Uploaded have been placed.",
+	"pages-using-files-desc": " A count of all non-duplicate main namespace pages, on all wikis, on which Uploaded Files have been placed.",
```
mooeypoo: This message has a space in the beginning -- is that on purpose? I don't think we can trust our translators to fully notice this, so if this is required for styling, we should add some spacing in the code between this message and whatever preceeds it.

Is this on purpose, or is it a fluke?

samwilson: Nope, that's a typo. :) Oops. Fixed.

PR #287, opened 2019-04-13
Always process some events via cronjob

MaxSem:
It just seems way more stable for slow jobs.

The numbers for cutoff are chosen based on statistics:
* Most events don't have a single category, so 1 category is already a lot
* There are ~30 participants per event on average
* Average event lasts for less than an hour (eh?)

Bug: T220463

### DIFF
--- a/src/AppBundle/Controller/EventDataController.php
+++ b/src/AppBundle/Controller/EventDataController.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
 
 namespace AppBundle\Controller;
 
+use AppBundle\Model\Event;
 use AppBundle\Model\Job;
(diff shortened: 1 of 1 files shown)


MusikAnimal:
I'm fine with this, but I think we should increase the rate of the cron. Do you see any issue with it running every minute? All it does is run the `app:spawn-jobs`command, and Event Metrics jobs are flagged as started in the db immediately, so we shouldn't get one cron job conflicting with another.

MaxSem:
Yeah, I've set it to run every minute on staging. In process, I noticed that jobs processing was missing from the crontab at all (prod is fine) - that would explain some of the shenanigans that happened.

PR #293, opened 2019-05-22
Start using the actors table

MaxSem:
Only a few uses are covered because this is just an attempt to organize
the actors support, however this change is self-contained and can be
merged on its own.

Bug: T219328

### DIFF
--- a/src/AppBundle/Repository/Repository.php
+++ b/src/AppBundle/Repository/Repository.php
@@ -15,6 +15,7 @@ use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityRepository;
+use PDO;
 use Psr\Cache\CacheItemPoolInterface;
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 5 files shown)


src/AppBundle/Repository/Repository.php
```diff
@@ -318,6 +320,25 @@ public function getUsernameFromId(int $userId): ?string
...
+    {
+        $conn = $this->getReplicaConnection();
+        $rqb = $conn->createQueryBuilder();
+
+        $rqb->select('actor_id')
+            ->from("$dbName.actor")
+            ->where('actor_name IN (:usernames)')
+            ->setParameter('usernames', $usernames, Connection::PARAM_STR_ARRAY);
+
+        return $this->executeQueryBuilder($rqb)->fetchAll(PDO::FETCH_FUNC, 'intval');
```
MusikAnimal: `ResultStatement::fetchAll()` seems to want type `int|null` as the second argument.

MaxSem: It documents the FETCH_FUNC usage, so the type annotation is wrong. In any case, there's a test that verify that it works as intended.

PR #294, opened 2019-05-24
Cache some wiki information in EventWikiRepository

MaxSem:


src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -22,6 +22,12 @@ class EventWikiRepository extends Repository
     /** Max number of pages to be returned when generating page IDs. */
     public const MAX_PAGES = 50000;
 
+    /** @var string[]|null[] */
+    private $domainsFromWiki = [];
+
+    /** @var array string[]|null[] */
```
MusikAnimal: Just need to remove the `array` to get the tests to pass

src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -91,10 +99,14 @@ public function getWikiFamilyName(string $value): ?string
     /**
      * Get the database name of the given (partial) domain name.
      * @param string $domain The domain name, without trailing '.org'.
-     * @return string Null if not found.
+     * @return string
```
MusikAnimal: The phpcs didn't complain for whatever reason, but I think this should be `string|null`

PR #297, opened 2019-05-30
Finalize the actors thingie

MaxSem:


src/AppBundle/Repository/EventWikiRepository.php
```diff
@@ -447,16 +447,16 @@ public function getSinglePageCreatedData(
...
-        $userRevisionTable = empty($usernames) ? 'revision' : 'revision_userindex';
+        $userRevisionTable = $actors ? 'revision' : 'revision_userindex';
 
         $sql = "SELECT `metric`, `value` FROM (
                     (
-                        SELECT 'creator' AS `metric`, rev_user_text AS `value`
-                        FROM $dbName.revision
-                        WHERE rev_page = :pageId
+                        SELECT 'creator' AS `metric`, actor_name AS `value`
+                        FROM $dbName.revision, $dbName.actor
```
MusikAnimal: I had to research this syntax, and apparently it's an older way to do JOINs. From what I'm reading ANSI-92 (newer) syntax is preferred. https://stackoverflow.com/a/1599201 What's your take? I realize this is an especially simple query

MusikAnimal:
I'm going to merge so we can get this out but I am curious about your thoughts on the comma syntax versus JOIN. I've never seen the comma variant until now!

src/AppBundle/Repository/EventWikiRepository.php
MaxSem: Mhm, I'm kinda used to implicit joins - LMK if you'd like me to make it explicit and I'll commit a followup.

PR #311, opened 2019-11-24
Remove the unnecessary word "success" from English messages

amire80:
See https://www.mediawiki.org/wiki/Localisation#Avoid_jargon_and_slang


MusikAnimal:
Thanks! Something is broken with our CI build, I'm just going to ignore it.

PR #315, opened 2021-02-25
Use ReplicasClient to access DBs

samwilson:


samwilson:
CI is failing because of replicas connectivity issues (they seem to be common from Travis). But try it locally, and see what you think.

There's a bit of stuff here that can be moved into a separate commit if required. The important parts of this patch are a) the passing of $dbName to `Repository:;getReplicaConnection()` (of course), and b) the way that `EventRepository::getRevisionsData()` now runs each query separately. This latter need more thought. We could UNION those that are to the same slice (I've not done so because it's more complicated and I want to see what y'all think).

MusikAnimal:
> There's a bit of stuff here that can be moved into a separate commit if required.

Up to you! As I said now that you can mark files as viewed, it's not so hard to review as one PR.

> the way that `EventRepository::getRevisionsData()` now runs each query separately. This latter need more thought. We could UNION those that are to the same slice (I've not done so because it's more complicated and I want to see what y'all think).

Yes! That's precisely what I did for XTools, see the [changes to GlobalContribsRepository](https://github.com/x-tools/xtools/commit/2e558be94ad12fb805d7ac87ee4256ae0e00b5b0#diff-2e5a1b0af30c97ec67958d5a708e639502c08f3afea6b4c095aea9235e57045b). If you want to do it that way, that's fine, or just query each db one by one.

You may or may not also run into the same issue I had with the pagination, since now you can't OFFSET the results and ensure you're not missing any data. I ended up ditching the page number links, and making the "Newer _N_" and "Older _N_" buttons change the `rev_timestamp >=` and `rev_timestamp <` clauses accordingly. See https://github.com/x-tools/xtools/blob/4f5fada21f8d10a7a8de488cb336c1d131f7325f/src/AppBundle/Repository/Repository.php#L350-L361. As I recall the /revisions page in Event Metrics wasn't a popular feature, so for now even removing pagination altogether is maybe acceptible (?), depending on the amount of energy you have to spare! I can also take a stab it in a separate PR, if you'd like.

src/AppBundle/Service/JobHandler.php
```diff
@@ -213,20 +219,23 @@ private function getQuota(): int
...
-        )->fetchColumn(0);
+        foreach ($registry->getConnections() as $conn) {
+            if (!$conn->isConnected()) {
+                continue;
+            }
+            $processCount += (int)$conn->executeQuery(
+                'SELECT COUNT(*) FROM information_schema.PROCESSLIST'
+            )->fetchFirstColumn();
+        }
+        return $processCount;
```
MusikAnimal, line 238: As I understand it, you can have 5 open connections to _any_ host, so this might incorrectly tell the system we're out of quota when we're not necessarily. I definitely, definitely do NOT think it's worth our time trying to figure out what connections we want to use when processing an event, and determining if there's quota only on those hosts, etc. The worse that will happen is it will force the event to be processed via the job queue, which is fine. Our low user base suggests we rarely if ever hit the connection limit anyway.

That said, I'm not sure about lines 231-233 That's skipping any connections we don't have an open connection to _in this process_, right? Do we know for sure that the connections we will need are already open at this point?

app/config/config.yml
```diff
@@ -62,27 +68,59 @@ doctrine:
...
+            toolforge_s7:
+                host: '%database.replica.host%'
+                port: '%database.replica.port_s7%'
+                user: '%database.replica.user%'
+                password: '%database.replica.password%'
+            toolforge_s8:
+                host: '%database.replica.host%'
+                port: '%database.replica.port_s8%'
+                user: '%database.replica.user%'
+                password: '%database.replica.password%'
```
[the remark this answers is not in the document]
samwilson: Fixed.

src/AppBundle/Service/JobHandler.php
samwilson: You're right though, the way it's done is not correct! I didn't think it through properly.

I'm not sure what the best way is then, because surely if we connect to the 8 slices to check the connection count, we'll already be over the permitted 5? (By the way, where's that documented? I haven't been able to find details on wikitech.) Or no, it'd close each as we go wouldn't it?

Could we just set it to 5 and see how it goes?

MusikAnimal: No I think we get 5 per host, so there's no risk in having connections open to all 8 simultaneously. Not sure if it's documented on Wikitech, and I can't even find it through codesearch either... I believe the limit is actually 10 (see old tasks like [T240036](https://phabricator.wikimedia.org/T240036) and [T177570](https://phabricator.wikimedia.org/T177570)), but we allow no more than 5 in Event Metrics to give some wiggle room for queries that aren't part of event processing.

This is pretty tricky, because we don't know which sections (aka slices) we need to connect to at this point in the code. It's a bit sloppy... but for now maybe just loop through s1 through s8 and return whichever has the highest value? That should effectively guard us from hitting the max connection limit, but at the cost of possibly making some jobs stay in a pending state longer than they need to (say s1 is full, but there's a job that only need s2).

samwilson: Okay, sounds good; I've done that.

MusikAnimal:
Looks like in order for the tests to pass, we need to change the doctrine.dbal.connections.eventmetrics.host var in config_test.yml to be `"%database.test.host%"`. Just `localhost` doesn't seem to work anymore (and doesn't appear to work on Travis, either).

samwilson (commented):
> Looks like in order for the tests to pass, we need to change the doctrine.dbal.connections.eventmetrics.host var in config_test.yml to be "%database.test.host%". Just localhost doesn't seem to work anymore (and doesn't appear to work on Travis, either).

Done.

PR #316, opened 2021-06-17
Bump Symfony core

MusikAnimal:
Bug: T284840

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3559,6 +3559,85 @@
             "time": "2021-01-07T16:49:33+00:00"
         },
         {
+            "name": "symfony/polyfill-php81",
+            "version": "v1.23.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/symfony/polyfill-php81.git",
+                "reference": "e66119f3de95efc359483f810c4c3e6436279436"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/symfony/polyfill-php81/zipball/e66119f3de95efc359483f810c4c3e6436279436",
+                "reference": "e66119f3de95efc359483f810c4c3e6436279436",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=7.1"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-main": "1.23-dev"
+                },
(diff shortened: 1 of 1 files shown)


sbassett29:
LGTM, thanks.

sbassett29:
LGTM, thanks.

Daimona:
I've received an email alert that might be related, with an error from the deploy script:

```
> php bin/console cache:clear
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ClassNotFoundException: Attempted to load class "DoctrineFixturesBundle" from namespace "Doctrine\Bundle\FixturesBundle".
Did you forget a "use" statement for another namespace? in /var/www/html/app/AppKernel.php:33
Stack trace:
#0 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(446): AppKernel->registerBundles()
#1 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(133): Symfony\Component\HttpKernel\Kernel->initializeBundles()
#2 /var/www/html/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php(169): Symfony\Component\HttpKernel\Kernel->boot()
#3 /var/www/html/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php(75): Symfony\Bundle\FrameworkBundle\Console\Application->registerCommands()
#4 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php(149): Symfony\Bundle\FrameworkBundle\Console\Application->doRun(Object(Symfo in /var/www/html/app/AppKernel.php on line 33
```

Is it something known? Also, it doesn't seem like it might've been caused by this specific commit. Perhaps `doctrine/doctrine-fixtures-bundle` shouldn't be a dev dependency if something from that library is used in actual code? It's always been under dev dependencies though.

MusikAnimal:
@Daimona No, not related. That's the ToolforgeBundle deploy script which [incorrectly assumes](https://github.com/wikimedia/ToolforgeBundle/blob/master/bin/deploy.sh#L87) that the app has Doctrine migrations. It doesn't actually cause any problems though, but we should still add a flag or some sort of try/catch.

PR #320, opened 2022-06-03
First pass at replacing the CI jobs with GitHub Actions

theresnotime:


MusikAnimal:
We're now using GitHub Actions

